PR #6, opened 2019-11-06
Rr/rpp simd optimizations

rrawther:
More simd functions added.
Added CMake option to enable/disable simd


src/imgaug/cpu/host_image_augmentations.hpp
```diff
@@ -1,6 +1,7 @@
 #include <cpu/rpp_cpu_common.hpp>
 #include <stdlib.h>
 #include <time.h>
+#include <omp.h>
```
Reza-Najafi: Why do we need this header here? #pragam omp ...  does not need it

rrawther: I just moved it from somewhere in the middle of the file, so we can include all the header files at one place in the beginning. If you don't need it, please remove it

Issue #29, opened 2020-05-28
CMakeList - MIOpen Dependencies


kiritigowda:
MIOpen Dependencies should be removed from RPP.

```
-- AMDGCN assembler: MIOPEN_AMDGCN_ASSEMBLER-NOTFOUND
```

kiritigowda:
@LokeshBonta can you look at this issue

LokeshBonta:
@kiritigowda  Created a PR 
[](https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/30)

kiritigowda:
#30 merged.

PR #30, opened 2020-05-28
Remove MIOPEN dependency in RPP build set-up

LokeshBonta:


LokeshBonta:
@kiritigowda 
Hi,
removed MiOpen dependency for RPP raised in [#29](https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/issues/29#event-3380592208)

PR #51, opened 2021-04-20
Enable static build option for logical_operations

r-abishek:


src/modules/hip/kernel/tensor.cpp
```diff
@@ -1,249 +1,379 @@
...
+    int id_z = hipBlockIdx_z * hipBlockDim_z + hipThreadIdx_z;
+
+    if (id_x >= a || id_y >= b || id_z >= c)
+    {
+        return;
+    }
+
     int pixIdx = id_y * c * a + id_x * c + id_z;
-    
     output[pixIdx] = (unsigned short)(input[pixIdx] * 257);
```
rrawther, line 226: What is the logic of this? Why is it multiplied by 257

r-abishek: @rrawther its used to convert a 0-255 range U8 to a 0-65535 range U16 (255 * 257 = 65535)

rrawther (approved):
@paveltc: please run all the unit-tests for this so we can merge

PR #53, opened 2021-04-21
Enable static build option for filter/morphological operations

r-abishek:


src/modules/hip/kernel/bilateral_filter.cpp
```diff
@@ -1,209 +1,207 @@
...
-
                 }
             }
 
-            int res1 = sum1/w_sum1;
-            int res2 = sum2/w_sum2;
-            int res3 = sum3/w_sum3;
-
+            int res1 = sum1 / w_sum1;
+            int res2 = sum2 / w_sum2;
```
rrawther, line 194: res2 and res3 shouldn't be computed unless channel==3

rrawther (approved):
@paveltc make sure the unit tests pass before merging

PR #54, opened 2021-04-22
Enable static build option for geometric/fused operations

r-abishek:


src/modules/hip/hip_declarations_inline.hpp
```diff
@@ -213,169 +43,15 @@ color_convert_hip_batch(T* srcPtr, U* dstPtr, RppiColorConvertMode convert_mode,
...
 
 template <typename T, typename U>
 RppStatus
 non_linear_blend_hip_batch_tensor(T* srcPtr1, T* srcPtr2, U* dstPtr, rpp::Handle &handle, RPPTensorFunctionMetaData &tensor_info)
 {
     int in_plnpkdind = getplnpkdind(tensor_info._in_format), out_plnpkdind = getplnpkdind(tensor_info._out_format);
-    InitHandle *handle_obj = handle.GetInitHandle();
     Rpp32u max_height, max_width;
-    max_size(handle_obj->mem.mgpu.csrcSize.height, handle_obj->mem.mgpu.csrcSize.width, handle.GetBatchSize(), &max_height, &max_width);
+    max_size(handle.GetInitHandle()->mem.mgpu.csrcSize.height, handle.GetInitHandle()->mem.mgpu.csrcSize.width, handle.GetBatchSize(), &max_height, &max_width);
```
rrawther, line 54: what is the reason for this change? Previous version might be more efficient

src/modules/hip/hip_fused_functions.cpp
```diff
@@ -48,12 +46,539 @@ color_twist_hip_batch(Rpp8u* srcPtr, Rpp8u* dstPtr, rpp::Handle& handle, RppiChn
...
+RppStatus
+color_twist_hip_batch_tensor_fp16(Rpp16f *srcPtr, Rpp16f *dstPtr, rpp::Handle &handle, RPPTensorFunctionMetaData &tensor_info)
+{
+    int in_plnpkdind = getplnpkdind(tensor_info._in_format);
+    int out_plnpkdind = getplnpkdind(tensor_info._out_format);
+    Rpp32u max_height, max_width;
+    max_size(handle.GetInitHandle()->mem.mgpu.csrcSize.height, handle.GetInitHandle()->mem.mgpu.csrcSize.width, handle.GetBatchSize(), &max_height, &max_width);
+
+#if defined (HIPRTC)
+
```
rrawther, line 103: It is better to remove empty code block

src/modules/hip/kernel/warp_perspective.cpp
```diff
@@ -1,137 +1,184 @@
...
+    int indextmp = 0;
+    int xc = id_x - (dest_width[id_z] >> 1);
+    int yc = id_y - (dest_height[id_z] >> 1);
+    int perspective_index = id_z * 9;
+
+    float k_float = (float)((perspective[perspective_index + 0] * xc) + (perspective[perspective_index + 1] * yc)) + perspective[perspective_index + 2];
+    float l_float = (float)((perspective[perspective_index + 3] * xc) + (perspective[perspective_index + 4] * yc)) + perspective[perspective_index + 5];
+    float m_float = (float)((perspective[perspective_index + 6] * xc) + (perspective[perspective_index + 7] * yc)) + perspective[perspective_index + 8];
+
+    int k = (int) ((k_float / m_float) + (source_width[id_z] >> 1));
```
rrawther, line 119: need to revisit this to avoid this division

src/modules/hip/hip_declarations_inline.hpp
r-abishek: This was done to maintain uniformity for the hip functions. Almost all hip functions were using handle.GetInitHandle() except just a few that I made the change for. Shall check for performance and make a common change in a subsequent PR if there is difference.

src/modules/hip/hip_fused_functions.cpp
r-abishek: As soon as this set of PRs go through, we'll be removing RTC entirely, so these HIPRTC blocks containing / not containing code will be deleted and only the hip_exec_* call below will be retained. (without the compile time switch)

src/modules/hip/kernel/warp_perspective.cpp
r-abishek: noted

asalmanp:
@r-abishek can you check and fix the Codacy issues for this PR?

rrawther (approved):

kiritigowda:
@r-abishek can you take a quick look at this warning, if not valid we can merge this - https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/54/checks?check_run_id=2405940439

kiritigowda (approved):

PR #55, opened 2021-04-23
Enable static build option for color model functions

r-abishek:


kiritigowda:
@r-abishek conflicts need to be resolved

src/modules/hip/kernel/hue.cpp
```diff
@@ -123,200 +174,325 @@ extern "C" __global__ void huergb_pkd(     unsigned char *input,
...
+        hsv = convert_one_pixel_to_hsv(pixel);
+        hsv.y *= sat[id_z];
 
-  if ((id_y >= yroi_begin[id_z]) && (id_y <= yroi_end[id_z]) &&
-    (id_x >= xroi_begin[id_z]) && (id_x <= xroi_end[id_z])) {
-    hsv = convert_one_pixel_to_hsv(pixel);
-    hsv.y *= sat[id_z];
-    if (hsv.y > 1.0) {
-      hsv.y = 1.0;
+        if (hsv.y > 1.0)
```
rrawther, line 324: better to use min and max instruction here to avoid conditional branch

rrawther (changes requested):
@pavel please make sure the tests pass before merging

src/modules/hip/kernel/hue.cpp
r-abishek: Modified to use HIP's fmaxf, fminf

r-abishek:
> @r-abishek conflicts need to be resolved

@kiritigowda Should be resolved now

asalmanp (approved):

pavel:
@rrawther Thanks for notifying me, but it does not seem that I'm involved in this project in any way.

https://blog.github.com/2011-03-23-mention-somebody-they-re-notified/

PR #60, opened 2021-05-11
HIPRTC phaseout - CMake cleanup

r-abishek:
HIPRTC phaseout - CMake cleanup PR#60:
- Change ReadMe instructions to use either of cmake -DBACKEND=OCL .. or cmake -DBACKEND=HIP .. and removes the DCOMPILE=HIPRTC/DCOMPILE=STATIC option.
- Clean up CMake to adhere to content needed for HIP STATIC build.
- Remove a few unnecessary files


CMakeLists.txt
```diff
@@ -151,6 +155,9 @@ enable_clang_tidy(
     EXTRA_ARGS
 
 )
+
```
asalmanp, line 158: can you remove ClangTidy in the above lines as well? we shouldn't need to include it.

asalmanp (changes requested):

CMakeLists.txt
r-abishek: Removed

r-abishek:
@asalmanp @rrawther Both Clang-Tidy and Cppcheck dependencies are removed. The gpu hack is removed from CMake too.

CMakeLists.txt
```diff
@@ -329,45 +196,45 @@ set(CPACK_DEBIAN_PACKAGE_RELEASE "local")
...
-  string (APPEND CPACK_RPM_PACKAGE_RELEASE "%{?dist}")
+if(PROC_RESULT EQUAL "0" AND NOT EVAL_RESULT STREQUAL "")
+    string (APPEND CPACK_RPM_PACKAGE_RELEASE "%{?dist}")
 endif()
 
 set(CPACK_PACKAGING_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
 
 set(CPACK_DEBIAN_PACKAGE_HOMEPAGE   ${CPACK_PACKAGE_HOMEPAGE})
-#set(CPACK_DEBIAN_PACKAGE_DEPENDS    "miopen-opencl") # Remove Explict requirements
+# set(CPACK_DEBIAN_PACKAGE_DEPENDS    "miopen-opencl") # Remove Explict requirements
```
asalmanp, line 217: remove the commented out lines (217 and 219) if it is not needed

asalmanp (changes requested):

rrawther (commented):
do we need to keep the binary file a.out in the repo. If this is an intermediate file, it is better to remove it from the repo.

CMakeLists.txt
r-abishek: Sure, removed now

r-abishek:
> do we need to keep the binary file a.out in the repo. If this is an intermediate file, it is better to remove it from the repo.

@rrawther No, its actually not needed and removed from the repo.

asalmanp (approved):

PR #70, opened 2021-08-31
Introduce tensor support

r-abishek:
- Adds all RPP framework headers and changes needed for tensor support.
- Adds all unit test and performance test support for brightness tensor.


src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -2078,6 +2079,26 @@ inline RppStatus custom_convolve_image_host(T* srcPtr, RppiSize srcSize, U* dstP
...
+    roiPtrImage->xywhROI.xy.y = roiPtrInput->ltrbROI.lt.y;
+    roiPtrImage->xywhROI.roiWidth = roiPtrInput->ltrbROI.rb.x - roiPtrInput->ltrbROI.lt.x + 1;
+    roiPtrImage->xywhROI.roiHeight = roiPtrInput->ltrbROI.rb.y - roiPtrInput->ltrbROI.lt.y + 1;
+
+    return RPP_SUCCESS;
+}
+
+inline RppStatus compute_roi_boundary_check_host(RpptROIPtr roiPtrImage, RpptROIPtr roiPtr, RpptROIPtr roiPtrDefault)
+{
+    roiPtr->xywhROI.xy.x = RPPMAX2(roiPtrDefault->xywhROI.xy.x, roiPtrImage->xywhROI.xy.x);
```
rrawther, line 2094: better to std::max here since compiler will generate the best code for the target platform

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -10,45 +10,397 @@
...
+    px3 = _mm_shuffle_epi8(px3, mask);    /* shuffle to get [R13|R14|R15|R16|G13|G14|G15|G16 || B13|B14|B15|B16|R17|G17|B17|R18] - Need R13-16, G13-16, B13-16 */ \
+    px4 = _mm_unpackhi_epi8(px0, zero);    /* unpack 8 hi-pixels of px0 */ \
+    px5 = _mm_unpackhi_epi8(px1, zero);    /* unpack 8 hi-pixels of px1 */ \
+    px6 = _mm_unpackhi_epi8(px2, zero);    /* unpack 8 hi-pixels of px2 */ \
+    px7 = _mm_unpackhi_epi8(px3, zero);    /* unpack 8 hi-pixels of px3 */ \
+    px0 = _mm_unpacklo_epi8(px0, zero);    /* unpack 8 lo-pixels of px0 */ \
+    px1 = _mm_unpacklo_epi8(px1, zero);    /* unpack 8 lo-pixels of px1 */ \
+    px2 = _mm_unpacklo_epi8(px2, zero);    /* unpack 8 lo-pixels of px2 */ \
+    px3 = _mm_unpacklo_epi8(px3, zero);    /* unpack 8 lo-pixels of px3 */ \
+    p0R = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px0, zero));    /* unpack 4 lo-pixels of px0 - Contains R01-04 */ \
```
rrawther, line 69: probably you can avoid unpacking by modifying the mask for _mm_shuffle_epi8.

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -10,45 +10,397 @@
...
+    p3G = _mm_cvtepi32_ps(_mm_unpackhi_epi16(px3, zero));    /* unpack 4 hi-pixels of px3 - Contains G13-16 */ \
+    p0B = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px4, zero));    /* unpack 4 lo-pixels of px4 - Contains B01-04 */ \
+    p1B = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px5, zero));    /* unpack 4 lo-pixels of px5 - Contains B05-08 */ \
+    p2B = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px6, zero));    /* unpack 4 lo-pixels of px6 - Contains B09-12 */ \
+    p3B = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px7, zero));    /* unpack 4 lo-pixels of px7 - Contains B13-16 */ \
+}
+
+#define RPP_STORE48_F32PLN3_TO_U8PLN3(dstPtrTempR, dstPtrTempG, dstPtrTempB) \
+{ \
+    px4 = _mm_cvtps_epi32(p0R);    /* convert to int32 for R */ \
```
rrawther, line 85: what is p0R, p1R..? Better to make #define generic with all arguments passed

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -10,45 +10,397 @@
...
+    p3G = _mm_cvtepi32_ps(_mm_unpackhi_epi16(px4, zero));    /* pixels 12-15 of original px1 containing 16 G values */ \
+    p0B = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px2, zero));    /* pixels 0-3 of original px1 containing 16 G values */ \
+    p1B = _mm_cvtepi32_ps(_mm_unpackhi_epi16(px2, zero));    /* pixels 4-7 of original px1 containing 16 G values */ \
+    p2B = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px5, zero));    /* pixels 8-11 of original px1 containing 16 G values */ \
+    p3B = _mm_cvtepi32_ps(_mm_unpackhi_epi16(px5, zero));    /* pixels 12-15 of original px1 containing 16 G values */ \
+}
+
+#define RPP_STORE48_F32PLN3_TO_U8PKD3(dstPtrTemp) \
+{ \
+    px4 = _mm_cvtps_epi32(p0R);    /* convert to int32 for R01-04 */ \
```
rrawther, line 138: what is p0R, p1R, .. etc?

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -10,45 +10,397 @@
...
+    px1 = _mm_cvtps_epi32(p1);    /* pixels 4-7 */ \
+    px2 = _mm_cvtps_epi32(p2);    /* pixels 8-11 */ \
+    px3 = _mm_cvtps_epi32(p3);    /* pixels 12-15 */ \
+    px0 = _mm_packus_epi32(px0, px1);    /* pixels 0-7 */ \
+    px1 = _mm_packus_epi32(px2, px3);    /* pixels 8-15 */ \
+    px0 = _mm_packus_epi16(px0, px1);    /* pixels 0-15 */ \
+    _mm_storeu_si128((__m128i *)dstPtrTemp, px0);    /* store pixels 0-15 */ \
+}
+
+#define RPP_LOAD12_F32PKD3_TO_F32PLN3(srcPtrTemp) \
```
rrawther, line 195: it is not a good practice to use variables from calling routine inside a macro.

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -10,45 +10,397 @@
...
+{ \
+    p0R = _mm_loadu_ps(srcPtrTemp); \
+    p0G = _mm_loadu_ps(srcPtrTemp + 3); \
+    p0B = _mm_loadu_ps(srcPtrTemp + 6); \
+    p0A = _mm_loadu_ps(srcPtrTemp + 9); \
+    _MM_TRANSPOSE4_PS (p0R, p0G, p0B, p0A); \
+}
+
+#define RPP_STORE12_F32PLN3_TO_F32PLN3(dstPtrTempR, dstPtrTempG, dstPtrTempB) \
```
rrawther, line 204: Please pass all arguments to macro

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -10,45 +10,397 @@
...
+    p1G = _mm_cvtepi32_ps(_mm_unpackhi_epi16(px1, zero));    /* pixels 4-7 of original px1 containing 16 G values */ \
+    p2G = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px4, zero));    /* pixels 8-11 of original px1 containing 16 G values */ \
+    p3G = _mm_cvtepi32_ps(_mm_unpackhi_epi16(px4, zero));    /* pixels 12-15 of original px1 containing 16 G values */ \
+    p0B = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px2, zero));    /* pixels 0-3 of original px1 containing 16 G values */ \
+    p1B = _mm_cvtepi32_ps(_mm_unpackhi_epi16(px2, zero));    /* pixels 4-7 of original px1 containing 16 G values */ \
+    p2B = _mm_cvtepi32_ps(_mm_unpacklo_epi16(px5, zero));    /* pixels 8-11 of original px1 containing 16 G values */ \
+    p3B = _mm_cvtepi32_ps(_mm_unpackhi_epi16(px5, zero));    /* pixels 12-15 of original px1 containing 16 G values */ \
+}
+
+#define RPP_STORE48_F32PLN3_TO_I8PKD3(dstPtrTemp) \
```
rrawther, line 332: it is bad practice to use macros which assumes parameters from calling routine

src/modules/cpu/host_tensor_augmentations.hpp
```diff
@@ -0,0 +1,1093 @@
...
+                roiPtrImage = &roiImage;
+                compute_xywh_from_ltrb_host(roiPtrInput, roiPtrImage);
+            }
+            else if (roiType == RpptRoiType::XYWH)
+            {
+                roiPtrImage = roiPtrInput;
+            }
+
+            roiPtr = &roi;
+            compute_roi_boundary_check_host(roiPtrImage, roiPtr, roiPtrDefault);
```
rrawther, line 590: do you really need to do this?

rrawther (changes requested):
please address the comments in the review

src/include/cpu/rpp_cpu_common.hpp

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: I have changed it now to take all arguments to completely avoid any ambiguity. (There were 12-16 arguments in a few of them, so I figured if RPP always uses px0/1/2/3 for __m128i and p0/1/2/3 for __m128 as a protocol, we could skip passing those arguments)

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: Done

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: Done

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: Done

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: Done

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: We do need to unpack to float to perform the computations though

src/include/hip/rpp_hip_common.hpp
```diff
@@ -72,4 +114,441 @@ inline RppStatus generate_gaussian_kernel_gpu(Rpp32f stdDev, Rpp32f* kernel, Rpp
...
+
+// F16 loads and stores without layout toggle (8 F16 pixels)
+
+__device__ __forceinline__ void rpp_hip_load8_and_unpack_to_float8(half *srcPtr, uint srcIdx, d_float8 *src_f8)
+{
+    d_half8 src_h8;
+    src_h8 = *((d_half8 *)(&srcPtr[srcIdx]));
+
+    *(float2 *)&(src_f8->x) = __half22float2(src_h8.x.x);
+    *((float2 *)&(src_f8->x) + 1) = __half22float2(src_h8.x.y);
```
asalmanp, line 244: don't use +1 here. First, create a temp float 4 and fill it based on __half22float2(src_h8.x.x)/__half22float2(src_h8.x.y) and the assign that float4 to the src_f8->x. do the same for src_f8->y

asalmanp (changes requested):

src/include/hip/rpp_hip_common.hpp
```diff
@@ -72,4 +114,441 @@ inline RppStatus generate_gaussian_kernel_gpu(Rpp32f stdDev, Rpp32f* kernel, Rpp
...
+    *(float2 *)&(src_f8->y) = __half22float2(src_h8.y.x);
+    *((float2 *)&(src_f8->y) + 1) = __half22float2(src_h8.y.y);
+}
+
+__device__ __forceinline__ void rpp_hip_pack_float8_and_store8(half *dstPtr, uint dstIdx, d_float8 *dst_f8)
+{
+    d_half8 dst_h8;
+
+    dst_h8.x.x = __float22half2_rn(*(float2 *)&(dst_f8->x));
+    dst_h8.x.y = __float22half2_rn(*((float2 *)&(dst_f8->x) + 1));
```
asalmanp, line 254: please try to avoid this +1 in your code. Use the x, y, z, and w components of the dst_f8->x.

src/include/hip/rpp_hip_common.hpp
```diff
@@ -72,4 +114,441 @@ inline RppStatus generate_gaussian_kernel_gpu(Rpp32f stdDev, Rpp32f* kernel, Rpp
...
+
+    src_f24->z.x = make_float4(rpp_hip_unpack2(src.x.x), rpp_hip_unpack1(src.x.y), rpp_hip_unpack0(src.y.x), rpp_hip_unpack3(src.y.x));
+    src_f24->z.y = make_float4(rpp_hip_unpack2(src.y.y), rpp_hip_unpack1(src.z.x), rpp_hip_unpack0(src.z.y), rpp_hip_unpack3(src.z.y));
+}
+
+// F32 loads with layout toggle PKD3 to PLN3 (24 F32 pixels)
+
+__device__ __forceinline__ void rpp_hip_load24_pkd3_and_unpack_to_float24_pln3(float *srcPtr, uint srcIdx, d_float24 *src_f24)
+{
+    src_f24->x.x.x = srcPtr[srcIdx];
```
asalmanp, line 281: this function needs to be rewritten. please don't access the srcPtr memory like the way you've written here by adding some offset. read a float4 each time and then use the x, y, z, w components to fill the src_f24.

asalmanp (changes requested):

src/include/hip/rpp_hip_common.hpp
```diff
@@ -72,4 +114,441 @@ inline RppStatus generate_gaussian_kernel_gpu(Rpp32f stdDev, Rpp32f* kernel, Rpp
...
+
+    src_f24->z.x = make_float4(rpp_hip_unpack2(src.x.x), rpp_hip_unpack1(src.x.y), rpp_hip_unpack0(src.y.x), rpp_hip_unpack3(src.y.x));
+    src_f24->z.y = make_float4(rpp_hip_unpack2(src.y.y), rpp_hip_unpack1(src.z.x), rpp_hip_unpack0(src.z.y), rpp_hip_unpack3(src.z.y));
+}
+
+// F16 loads with layout toggle PKD3 to PLN3 (24 F16 pixels)
+
+__device__ __forceinline__ void rpp_hip_load24_pkd3_and_unpack_to_float24_pln3(half *srcPtr, uint srcIdx, d_float24 *src_f24)
+{
+    src_f24->x.x.x = __half2float(srcPtr[srcIdx]);
```
asalmanp, line 329: use the same comments mentioned above for rpp_hip_load24_pkd3_and_unpack_to_float24_pln3 function to rewrite this function as well.

src/include/hip/rpp_hip_common.hpp
```diff
@@ -72,4 +114,441 @@ inline RppStatus generate_gaussian_kernel_gpu(Rpp32f stdDev, Rpp32f* kernel, Rpp
...
+// F32 loads with layout toggle PLN3 to PKD3 (24 F32 pixels)
+
+__device__ __forceinline__ void rpp_hip_load24_pln3_and_unpack_to_float24_pkd3(float *srcPtr, uint srcIdx, uint increment, d_float24 *src_f24)
+{
+    float *srcPtrR, *srcPtrG, *srcPtrB;
+    srcPtrR = srcPtr + srcIdx;
+    srcPtrG = srcPtrR + increment;
+    srcPtrB = srcPtrG + increment;
+
+    src_f24->x.x.x = *srcPtrR;
```
asalmanp, line 386: use vector float4 for reading srcPtr memory and use its components when filling the src_f24.

asalmanp (changes requested):

include/rppdefs.h
```diff
@@ -22,28 +29,28 @@ typedef unsigned long long  Rpp64u;
...
 
- typedef enum
- {
-    rppStatusSuccess = 0,
-    rppStatusBadParm   = 1,
-    rppStatusUnknownError = 2,
+typedef enum
+{
+    rppStatusSuccess        = 0,
+    rppStatusBadParm        = 1,
```
rrawther, line 44: it is better to give -ve values for error codes

src/include/cpu/rpp_cpu_simd.hpp
rrawther: px0 through px7 is reassigned inside the macro. In that case, please avoid passing them through macro and declare as local variables inside. I think in this case it is better to make it as inline function instead.
You can have a common load macro which would take a "function name" and variable number of arguments, which can call different load routines inside the macro, so the calling macro will be the same.

rrawther (changes requested):
please redo the macros and fix codacy warnings

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: Done. I have added this cleaner handling with variadic inline functions and parameter pack templating now. Added more info on the email.

src/include/hip/rpp_hip_common.hpp

src/include/hip/rpp_hip_common.hpp

src/include/hip/rpp_hip_common.hpp

src/include/hip/rpp_hip_common.hpp

src/include/hip/rpp_hip_common.hpp

include/rppdefs.h
r-abishek: Done

rrawther (approved):
@paveltc : Please run all the unit-tests before merge

r-abishek:
@asalmanp The I8 images now match exactly with the F16/F32/U8 visually. Please let me know if you have any other thoughts too. We could now proceed for the final  testing and merge @paveltc

src/modules/hip/kernel/brightness.hpp
```diff
@@ -0,0 +1,288 @@
...
+
+__device__ void brightness_hip_compute(float *srcPtr, d_float8 *src_f8, d_float8 *dst_f8, float4 *alpha_f4, float4 *beta_f4)
+{
+    dst_f8->x = src_f8->x * *alpha_f4 + *beta_f4 * (float4)0.0039216;
+    dst_f8->y = src_f8->y * *alpha_f4 + *beta_f4 * (float4)0.0039216;
+}
+
+__device__ void brightness_hip_compute(signed char *srcPtr, d_float8 *src_f8, d_float8 *dst_f8, float4 *alpha_f4, float4 *beta_f4)
+{
+    dst_f8->x = rpp_hip_pixel_check((src_f8->x + 128) * *alpha_f4 + *beta_f4) - 128;
```
asalmanp, line 18: src_f8->x is a float4 so if your intention is to add each of the x/y/z/w with 128 then you need to use (float4)128 here.

similarly, you would need to use (float4)128 to subtract in the end of this line.

asalmanp (changes requested):

src/modules/hip/kernel/roi_conversion.hpp
```diff
@@ -0,0 +1,29 @@
+#include <hip/hip_runtime.h>
+
+extern "C" __global__ void roi_converison_ltrb_to_xywh(int *roiTensorPtrSrc)
+{
+    int id_x = (hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x) * 4;
+
+    roiTensorPtrSrc[id_x + 2] -= (roiTensorPtrSrc[id_x] - 1);
```
asalmanp, line 7: use int2 to read from the roiTensorPtrSrc instead of using +1

asalmanp (changes requested):

src/modules/rppi_validate.hpp
```diff
@@ -12,6 +12,93 @@
...
+        {
+            layoutParams.channelParam = 1;
+            layoutParams.bufferMultiplier = 3;
+        }
+    }
+
+    return layoutParams;
+}
+
+// inline void copy_roiTensor(RpptROIPtr roiTensorPtrSrc, rpp::Handle& handle)
```
asalmanp, line 43: remove commented out codes if it is not needed.

src/include/hip/rpp_hip_common.hpp
```diff
@@ -72,4 +132,488 @@ inline RppStatus generate_gaussian_kernel_gpu(Rpp32f stdDev, Rpp32f* kernel, Rpp
...
+__device__ __forceinline__ float4 rpp_hip_unpack(uint src)
+{
+    return make_float4(rpp_hip_unpack0(src), rpp_hip_unpack1(src), rpp_hip_unpack2(src), rpp_hip_unpack3(src));
+}
+
+// Un-Packing from I8s
+
+__device__ __forceinline__ float rpp_hip_unpack0(int src)
+{
+    return (float)(signed char)(src & 0xFF);
```
asalmanp, line 196: why the casting to (signed char) is needed here?

asalmanp:
@r-abishek please take a look at my latest comments and make the necessary changes.

src/modules/hip/kernel/brightness.hpp
r-abishek: Thats true, but I believe it was doing an auto typecast to float4. In any case, I have added a manual typecast now to be on the safer side.

src/modules/hip/kernel/roi_conversion.hpp
r-abishek: Thanks! I have changed it now. Also, here roiTensorPtrSrc has 0/1/2/3, so I have used int4.

src/modules/rppi_validate.hpp
r-abishek: Done. Cleaned up this rppi_validate file for just some formatting changes, tabs to spaces, and removal of unwanted commented code.

src/include/hip/rpp_hip_common.hpp
r-abishek: Actually I found this to be the only way to do the unpack correctly. It takes  src, gets the first byte by masking with 0xFF, then we explicitly tell the compiler to interpret the MSB as a signed bit, then convert to float. If I directly convert to float, the most significant bit was interpreted not as a sign, and as part of an 8 bit number.

asalmanp (approved):

rrawther:
@paveltc please run rali and rpp unit-tests on this PR so we can merge it

paveltc:
@rrawther @asalmanp This PR passes the unit tests.

r-abishek:
> @rrawther @asalmanp This PR passes the unit tests.

@asalmanp @kiritigowda Could we merge the PR#70 too? The unit tests have been verified.

PR #78, opened 2021-10-27
Vector strides for hip kernels

r-abishek:
@rrawther @kiritigowda This is a minor PR that replicates the passing of vector strides mod (using uint2 / uint3 for strides). The PR makes the same change for the previous 3 tensor functions - brightness, blend, gamma_correction.


paveltc:
@rrawther @kiritigowda @r-abishek This PR passes all unit tests.

PR #84, opened 2021-12-17
Gridmask augmentation tensor implementation

r-abishek:
@rrawther @asalmanp This PR adds Gridmask to RPP.

- Adds gridmask augmentation on host and hip
- Adds the corresponding test suite on host and hip for both unit and performance tests

### DIFF
--- a/src/include/cpu/rpp_cpu_common.hpp
+++ b/src/include/cpu/rpp_cpu_common.hpp
@@ -2066,21 +2066,6 @@ inline RppStatus custom_convolve_image_host(T* srcPtr, RppiSize srcSize, U* dstP
     return RPP_SUCCESS;
 }
 
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
 // Compute Functions for RPP Tensor API
 
 inline RppStatus compute_gridmask_masks_16_host(__m128 *pCol, __m128 *pGridRowRatio, __m128 pCosRatio, __m128 pSinRatio, __m128 pGridRatio, __m128 *pMask)
@@ -2523,18 +2508,6 @@ inline RppStatus compute_color_jitter_12_host(__m128 *p, __m128 *pCtm)
 
     return RPP_SUCCESS;
 }
-
-
-
-
-
-
-
-
-
-
-
-
 
 // Compute Functions for RPP Image API
 
--- a/src/include/cpu/rpp_cpu_common.hpp
+++ b/src/include/cpu/rpp_cpu_common.hpp
@@ -2081,7 +2081,71 @@ inline RppStatus custom_convolve_image_host(T* srcPtr, RppiSize srcSize, U* dstP
 
 
 
-// Compute Functions
+// Compute Functions for RPP Tensor API
+
+inline RppStatus compute_gridmask_masks_16_host(__m128 *pCol, __m128 *pGridRowRatio, __m128 pCosRatio, __m128 pSinRatio, __m128 pGridRatio, __m128 *pMask)
+{
+    __m128 pCalc[2];
+
+    pCalc[0] = _mm_fmadd_ps(pCol[0], pCosRatio, pGridRowRatio[0]);
+    pCalc[1] = _mm_fmadd_ps(pCol[0], pSinRatio, pGridRowRatio[1]);
+    pCalc[0] = _mm_cmpge_ps(_mm_sub_ps(pCalc[0], _mm_floor_ps(pCalc[0])), pGridRatio);
+    pCalc[1] = _mm_cmpge_ps(_mm_sub_ps(pCalc[1], _mm_floor_ps(pCalc[1])), pGridRatio);
+    pMask[0] = _mm_or_ps(pCalc[0], pCalc[1]);
+
+    pCalc[0] = _mm_fmadd_ps(pCol[1], pCosRatio, pGridRowRatio[0]);
+    pCalc[1] = _mm_fmadd_ps(pCol[1], pSinRatio, pGridRowRatio[1]);
+    pCalc[0] = _mm_cmpge_ps(_mm_sub_ps(pCalc[0], _mm_floor_ps(pCalc[0])), pGridRatio);
+    pCalc[1] = _mm_cmpge_ps(_mm_sub_ps(pCalc[1], _mm_floor_ps(pCalc[1])), pGridRatio);
+    pMask[1] = _mm_or_ps(pCalc[0], pCalc[1]);
+
+    pCalc[0] = _mm_fmadd_ps(pCol[2], pCosRatio, pGridRowRatio[0]);
+    pCalc[1] = _mm_fmadd_ps(pCol[2], pSinRatio, pGridRowRatio[1]);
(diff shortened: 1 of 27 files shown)


paveltc:
@rrawther @asalmanp @r-abishek This PR passes all of the unit tests.

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -2081,7 +2081,101 @@ inline RppStatus custom_convolve_image_host(T* srcPtr, RppiSize srcSize, U* dstP
...
-// Compute Functions
+// Compute Functions for RPP Tensor API
+
+inline RppStatus compute_gridmask_masks_16_host(__m128 *pCol, __m128 *pGridRowRatio, __m128 pCosRatio, __m128 pSinRatio, __m128 pGridRatio, __m128 *pMask)
+{
+    __m128 pCalc[2];
+
+    pCalc[0] = _mm_fmadd_ps(pCol[0], pCosRatio, pGridRowRatio[0]);
+    pCalc[1] = _mm_fmadd_ps(pCol[0], pSinRatio, pGridRowRatio[1]);
+    pCalc[0] = _mm_cmpge_ps(_mm_sub_ps(pCalc[0], _mm_floor_ps(pCalc[0])), pGridRatio);
```
rrawther, line 2092: Is this subtracting the rounded int value? Then use _mm_round_ps() instead

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -2081,7 +2081,101 @@ inline RppStatus custom_convolve_image_host(T* srcPtr, RppiSize srcSize, U* dstP
...
+    pCalc[1] = _mm_cmpge_ps(_mm_sub_ps(pCalc[1], _mm_floor_ps(pCalc[1])), pGridRatio);
+    pMask = _mm_or_ps(pCalc[0], pCalc[1]);
+    pCol = _mm_add_ps(pCol, xmm_p4);
+
+    return RPP_SUCCESS;
+}
+
+inline RppStatus compute_gridmask_result_48_host(__m128 *p, __m128 *pMask)
+{
+    p[0] = _mm_and_ps(p[0], pMask[0]);
```
rrawther, line 2138: better to do in AVX

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -2430,6 +2524,20 @@ inline RppStatus compute_color_jitter_12_host(__m128 *p, __m128 *pCtm)
     return RPP_SUCCESS;
 }
 
+
```
rrawther, line 2527: please remove empty lines

src/include/hip/rpp_hip_common.hpp
```diff
@@ -81,6 +87,12 @@ enum class RPPTensorDataType
 
 // schar
 typedef signed char schar;
+typedef struct d_schar3
```
rrawther, line 90: signed char is same as char. Can't we just use char3 built in type instead?

src/modules/hip/kernel/gridmask.hpp
```diff
@@ -0,0 +1,367 @@
...
+    }
+}
+
+// PLN3 -> PKD3
+template <typename T>
+__device__ void gridmask_result_pln3_pkd3_hip_compute(T *srcPtr, int srcIdx, uint3 *srcStridesNCH, T *dstPtr, int dstIdx, float2 *gridColRatio, float gridRatio)
+{
+    if ((gridColRatio->x - floorf(gridColRatio->x) >= gridRatio) || (gridColRatio->y - floorf(gridColRatio->y) >= gridRatio))
+    {
+        dstPtr[dstIdx] = srcPtr[srcIdx];
```
rrawther, line 139: can't you use vector load here?

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pkd3.cpp
```diff
@@ -968,6 +971,64 @@ int main(int argc, char **argv)
...
+            start = clock();
+
+            if (ip_bitDepth == 0)
+                rppt_gridmask_gpu(d_input, srcDescPtr, d_output, dstDescPtr, tileWidth, gridRatio, gridAngle, translateVector, d_roiTensorPtrSrc, roiTypeSrc, handle);
+            else if (ip_bitDepth == 1)
+                rppt_gridmask_gpu(d_inputf16, srcDescPtr, d_outputf16, dstDescPtr, tileWidth, gridRatio, gridAngle, translateVector, d_roiTensorPtrSrc, roiTypeSrc, handle);
+            else if (ip_bitDepth == 2)
+                rppt_gridmask_gpu(d_inputf32, srcDescPtr, d_outputf32, dstDescPtr, tileWidth, gridRatio, gridAngle, translateVector, d_roiTensorPtrSrc, roiTypeSrc, handle);
+            else if (ip_bitDepth == 3)
+                missingFuncFlag = 1;
```
rrawther, line 1016: are we going to support this missing functionality in future?

rrawther (changes requested):
@asalmanp : Please review HIP kernels for this PR

src/include/cpu/rpp_cpu_common.hpp
r-abishek: Done

src/include/cpu/rpp_cpu_common.hpp

src/modules/hip/kernel/gridmask.hpp
r-abishek: Done

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pkd3.cpp
r-abishek: Actually no, These other cases can be removed as part of a test suite change PR since it will affect the BatchPD/Tensor automation script too.

src/include/cpu/rpp_cpu_common.hpp
r-abishek: A lot of the host implementation for gridmask was before the avx discussion. I'll probably issue a separate PR to add the AVX version since SSE was already implemented. Wanted to merge this soon since the ricap implementations were dependent on the merge to avoid conflicts.

src/include/hip/rpp_hip_common.hpp
r-abishek: That's true, but looks like it might be compiler dependent too - in the way they interpret it. 
So Rpp32s was chosen to be signed char at the top level rpp_defs, and that ends up following down to the low level API.

std::is_same<int, int>::value = 1
std::is_same<int, signed int>::value = 1 // implies compiler interpreted int is same as signed int
std::is_same<int, unsigned int>::value = 0 // implies compiler interpreted int is different from unsigned int

std::is_same<char, char>::value = 1
std::is_same<char, signed char>::value = 0 // implies compiler interpreted char is different from signed char
std::is_same<char, unsigned char>::value = 0 // implies compiler interpreted char is different from unsigned char

asalmanp (approved):

rrawther (approved):

PR #86, opened 2022-01-06
License Updates

kiritigowda:
License Updated to 2022

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -18,7 +18,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'AMD-RPP'
-copyright = '2020, Lokeswara Rao Bonta(MCW-AMD)'
+copyright = '2020 - 2022, Lokeswara Rao Bonta(MCW-AMD)'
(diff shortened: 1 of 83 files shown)


docs/source/conf.py
```diff
@@ -18,7 +18,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'AMD-RPP'
-copyright = '2020, Lokeswara Rao Bonta(MCW-AMD)'
+copyright = '2020 - 2022, Lokeswara Rao Bonta(MCW-AMD)'
```
LakshmiKumar23, line 21: @kiritigowda This is different than the other copyrights. Is that okay? Seems like it's the only occurance though

LakshmiKumar23 (changes requested):

docs/source/conf.py
kiritigowda: Should be ok

LakshmiKumar23: Can be merged then

PR #87, opened 2022-01-18
Spatter augmentation tensor implementation

r-abishek:
@rrawther @asalmanp This PR adds Spatter augmentation to RPP.

Adds host and hip implementations for spatter augmentation.
Adds the corresponding test suite on host and hip for running both unit and performance tests.


src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -2068,6 +2068,86 @@ inline RppStatus custom_convolve_image_host(T* srcPtr, RppiSize srcSize, U* dstP
...
+    p[2] = _mm256_fmadd_ps(p[2], pSpatterMaskInv[0], _mm256_mul_ps(pSpatterValue[1], pSpatterMask[0]));    // spatter adjustment
+    p[3] = _mm256_fmadd_ps(p[3], pSpatterMaskInv[1], _mm256_mul_ps(pSpatterValue[1], pSpatterMask[1]));    // spatter adjustment
+    p[4] = _mm256_fmadd_ps(p[4], pSpatterMaskInv[0], _mm256_mul_ps(pSpatterValue[2], pSpatterMask[0]));    // spatter adjustment
+    p[5] = _mm256_fmadd_ps(p[5], pSpatterMaskInv[1], _mm256_mul_ps(pSpatterValue[2], pSpatterMask[1]));    // spatter adjustment
+
+    return RPP_SUCCESS;
+}
+
+inline RppStatus compute_spatter_24_host(__m256 *p, __m256 *pSpatterMaskInv, __m256 *pSpatterMask, __m256 *pSpatterValue)
+{
```
rrawther, line 2084: please check if pSpatterMaskInv is same for all

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -782,6 +792,65 @@ inline RppStatus rpp_store48_f32pln3_to_u8pkd3_avx(Rpp8u *dstPtr, __m256 *p)
     return RPP_SUCCESS;
 }
 
+inline RppStatus rpp_load16_u8_to_f32_avx(Rpp8u *srcPtr, __m256 *p)
+{
+    __m128i px;
+    __m128i pxMask0 = _mm_setr_epi8(0, 0x80, 0x80, 0x80, 1, 0x80, 0x80, 0x80, 2, 0x80, 0x80, 0x80, 3, 0x80, 0x80, 0x80);
```
rrawther, line 798: keep these masks as constant globals since it is repeatedly used

src/modules/cpu/host_tensor_augmentations.hpp
```diff
@@ -9191,4 +9193,1581 @@ RppStatus gridmask_i8_i8_host_tensor(Rpp8s *srcPtr,
...
+        std::mt19937 gen(rd()); // Seeding rd() to fast mersenne twister engine
+        std::uniform_int_distribution<> distribX(0, 1920 - roiPtr->xywhROI.roiWidth);
+        std::uniform_int_distribution<> distribY(0, 1080 - roiPtr->xywhROI.roiHeight);
+
+        RppiPoint maskLoc;
+        maskLoc.x = distribX(gen);
+        maskLoc.y = distribY(gen);
+
+        Rpp32f *spatterMaskPtr, *spatterMaskInvPtr;
+        spatterMaskPtr = &spatterMask[(1920 * maskLoc.y) + maskLoc.x];
```
rrawther, line 9272: What is the difference between the spatterMask and InvSpatterMask

src/modules/cpu/host_tensor_augmentations.hpp
```diff
@@ -9191,4 +9193,1581 @@ RppStatus gridmask_i8_i8_host_tensor(Rpp8s *srcPtr,
...
+
+        Rpp32u bufferLength = roiPtr->xywhROI.roiWidth * layoutParams.bufferMultiplier;
+
+        Rpp8u *srcPtrChannel, *dstPtrChannel;
+        srcPtrChannel = srcPtrImage + (roiPtr->xywhROI.xy.y * srcDescPtr->strides.hStride) + (roiPtr->xywhROI.xy.x * layoutParams.bufferMultiplier);
+        dstPtrChannel = dstPtrImage;
+
+        std::random_device rd;  // Random number engine seed
+        std::mt19937 gen(rd()); // Seeding rd() to fast mersenne twister engine
+        std::uniform_int_distribution<> distribX(0, 1920 - roiPtr->xywhROI.roiWidth);
```
rrawther, line 9264: Does this work for image sizes more than 1920x1080

src/modules/hip/kernel/spatter.hpp
```diff
@@ -0,0 +1,322 @@
+#include <hip/hip_runtime.h>
+#include "hip/rpp_hip_common.hpp"
+#include "func_specific/spatter_mask.hpp"
+#include <random>
+
+__device__ void spatter_hip_compute(uchar *srcPtr, d_float8 *src_f8, d_float8 *dst_f8, d_float8 *mask_f8, d_float8 *maskInv_f8, float4 *pix_f4)
```
rrawther, line 6: can't you avoid different definitions using Template argument for the srcPtr?

rrawther (changes requested):

src/include/cpu/rpp_cpu_common.hpp
r-abishek: Yes, the spatterMask/spatterMaskInv only contains the mask for the spattered regions in the image. So they should be the same for the R/G/B channels.

src/modules/cpu/host_tensor_augmentations.hpp
r-abishek: The spatterMask is basically 1 - spatterMaskInv and both are needed. Instead of computing that at runtime, we have both statically initialized.

src/modules/cpu/host_tensor_augmentations.hpp
r-abishek: Actually no. Had planned to have the max as 4K (3840 x 2160), but the static declaration becomes too large, so decided to cap it at 1920 x 1080. I have mentioned it in the header documentation -> https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/87/files/ebc88e3851dc6ea7c3b723eb3659667717b61c1a#diff-72f99bf4cde5647f79cb11784b92bfa86370aedb5fb7ebc9a8b114cd4405c988R237

src/modules/hip/kernel/spatter.hpp
r-abishek: Templating could be used if every arithmetic inside the 4 functions are exactly the same except for the data-type, which can be templated as <T>. But in this case, we have different computations for each function.

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: Done

src/modules/hip/kernel/spatter.hpp
```diff
@@ -0,0 +1,322 @@
...
+
+__device__ void spatter_hip_compute(schar *srcPtr, d_float8 *src_f8, d_float8 *dst_f8, d_float8 *mask_f8, d_float8 *maskInv_f8, float4 *pix_f4)
+{
+    dst_f8->x = ((src_f8->x + (float4)128) * maskInv_f8->x) + (*pix_f4 * mask_f8->x) - (float4)128;
+    dst_f8->y = ((src_f8->y + (float4)128) * maskInv_f8->y) + (*pix_f4 * mask_f8->y) - (float4)128;
+}
+
+__device__ void spatter_hip_compute(half *srcPtr, d_float8 *src_f8, d_float8 *dst_f8, d_float8 *mask_f8, d_float8 *maskInv_f8, float4 *pix_f4)
+{
+    float4 pixNorm_f4 = *pix_f4 * (float4)0.0039216;
```
asalmanp, line 27: what's this 0.0039216 magic number? can you add a comment for every const number used in these kernels?

asalmanp (changes requested):

src/modules/hip/kernel/spatter.hpp
```diff
@@ -0,0 +1,322 @@
...
+    }
+    else if (dstDescPtr->c == 1)
+    {
+        float meanGreyVal = ((float)spatterColor.B + (float)spatterColor.G + (float)spatterColor.R) * 0.3333;
+        spatterColor_f3 = make_float3(meanGreyVal, meanGreyVal, meanGreyVal);
+    }
+
+    float *spatterMaskPtr, *spatterMaskInvPtr;
+    Rpp32u maskSize = 2073600 * sizeof(float);
+    hipMalloc(&spatterMaskPtr, maskSize);
```
asalmanp, line 238: you should not allocate memory here during the launch, the allocation should occur outside of this function and pass a pointer to this hip_exec_spatter_tensor. Please remove line 238-241 from here.

src/modules/hip/kernel/spatter.hpp
```diff
@@ -0,0 +1,322 @@
...
+                               make_uint2(dstDescPtr->strides.nStride, dstDescPtr->strides.hStride),
+                               spatterMaskPtr,
+                               spatterMaskInvPtr,
+                               maskLocArr,
+                               spatterColor_f3,
+                               roiTensorPtrSrc);
+        }
+    }
+
+    hipFree(&spatterMaskPtr);
```
asalmanp, line 318: same comment as above. freeing a memory here is not correct and safe. freeing memory for the masks should be outside of this function.

asalmanp (changes requested):

src/modules/rppt_tensor_augmentations.cpp
```diff
@@ -1180,3 +1180,171 @@ rppt_gridmask_host(RppPtr_t srcPtr,
...
+    uint2 maskLocArrHost[dstDescPtr->n];
+    for(int i = 0; i < dstDescPtr->n; i++)
+    {
+        std::uniform_int_distribution<> distribX(0, SPATTER_MAX_WIDTH - roiTensorPtrSrcHost[i].xywhROI.roiWidth);
+        std::uniform_int_distribution<> distribY(0, SPATTER_MAX_HEIGHT - roiTensorPtrSrcHost[i].xywhROI.roiHeight);
+        maskLocArrHost[i].x = distribX(gen);
+        maskLocArrHost[i].y = distribY(gen);
+    }
+    uint2 *maskLocArr;
+    hipMalloc(&maskLocArr, dstDescPtr->n * sizeof(uint2));
```
asalmanp, line 1223: same comment as above. please remove allocation from this function.

src/modules/rppt_tensor_augmentations.cpp
```diff
@@ -1180,3 +1180,171 @@ rppt_gridmask_host(RppPtr_t srcPtr,
...
+                           srcDescPtr,
+                           static_cast<Rpp8s*>(dstPtr) + dstDescPtr->offsetInBytes,
+                           dstDescPtr,
+                           maskLocArr,
+                           spatterColor,
+                           roiTensorPtrSrc,
+                           roiType,
+                           rpp::deref(rppHandle));
+    }
+    hipFree(&maskLocArr);
```
asalmanp, line 1274: move hipFree outside of this function.

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pkd3.cpp
```diff
@@ -1029,6 +1032,74 @@ int main(int argc, char **argv)
...
+            // spatterColor.R = 98;
+            // spatterColor.G = 3;
+            // spatterColor.B = 3;
+
+            // Ink Spatter
+            // spatterColor.R = 5;
+            // spatterColor.G = 20;
+            // spatterColor.B = 64;
+
+            // Uncomment to run test case with an xywhROI override
```
asalmanp, line 1056: we talked about these comments before that you would need to pass these as parameters to the test. When are you going to make these changes?

asalmanp (changes requested):

src/modules/hip/kernel/spatter.hpp
r-abishek: Done. #define is now used to initialize a common 1/255

src/modules/hip/kernel/spatter.hpp
r-abishek: I've moved the allocation out. Now pre-allocating temp hip buffers like for other functions.
I presume copying new data into a pre-allocated buffer is okay to have in the call path?

src/modules/hip/kernel/spatter.hpp
r-abishek: Done

src/modules/rppt_tensor_augmentations.cpp
r-abishek: Done

src/modules/rppt_tensor_augmentations.cpp
r-abishek: Done

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pkd3.cpp
r-abishek: Since every function in the test suite has a different number and type of parameters taken from the user, this change might not actually be a direct one. Some functions take a single u8/uint/f32 or some functions take whole arrays passed in, and others have tensor based ROIs. Will take a look on this in a separate test suite PR. Should be after the next couple of essential PRs ricap/resize/colortwist/cmn are merged.

src/modules/cpu/host_tensor_augmentations.hpp
```diff
@@ -9261,16 +9261,16 @@ RppStatus spatter_u8_u8_host_tensor(Rpp8u *srcPtr,
 
         std::random_device rd;  // Random number engine seed
         std::mt19937 gen(rd()); // Seeding rd() to fast mersenne twister engine
-        std::uniform_int_distribution<> distribX(0, 1920 - roiPtr->xywhROI.roiWidth);
-        std::uniform_int_distribution<> distribY(0, 1080 - roiPtr->xywhROI.roiHeight);
+        std::uniform_int_distribution<> distribX(0, SPATTER_MAX_WIDTH - roiPtr->xywhROI.roiWidth);
```
rrawther, line 9264: Is SPATTER_MAX_WIDTH and height user parameters?

src/modules/hip/handlehip.cpp
```diff
@@ -247,6 +247,7 @@ struct HandleImpl
             hipMalloc(&(this->initHandle->mem.mgpu.charArr[i].charmem) , sizeof(Rpp8s) * this->nBatchSize);
 	    }
         hipMalloc(&(this->initHandle->mem.mgpu.rgbArr.rgbmem) , sizeof(RpptRGB) * this->nBatchSize);
+        hipMalloc(&(this->initHandle->mem.mgpu.maskArr.floatmem) , sizeof(Rpp32f) * 8294400);    // 3840 x 2160
```
rrawther, line 250: it is better not name the array with .floatmem (is this a union?). Please keep it simple.

rrawther (approved):
Can be merged after @asalmanp approves HIP changes

src/modules/cpu/host_tensor_augmentations.hpp
r-abishek: No, the SPATTER_MAX_WIDTH is defined at compile time. 1920 for now.

src/modules/hip/handlehip.cpp
r-abishek: This one actually was developed previously for rpphandle from batchpd. Just used it here. Shall look into if changes can be made in a subsequent pr.

src/modules/hip/kernel/spatter.hpp
```diff
@@ -0,0 +1,327 @@
...
+        float meanGreyVal = ((float)spatterColor.B + (float)spatterColor.G + (float)spatterColor.R) * 0.3333;
+        spatterColor_f3 = make_float3(meanGreyVal, meanGreyVal, meanGreyVal);
+    }
+
+    Rpp32u maskSize = SPATTER_MAX_WIDTH * SPATTER_MAX_HEIGHT;
+    Rpp32u maskSizeFloat = maskSize * sizeof(float);
+    float *spatterMaskPtr, *spatterMaskInvPtr;
+    spatterMaskPtr = handle.GetInitHandle()->mem.mgpu.maskArr.floatmem;
+    spatterMaskInvPtr = handle.GetInitHandle()->mem.mgpu.maskArr.floatmem + maskSize;
+    hipMemcpy(spatterMaskPtr, spatterMask, maskSizeFloat, hipMemcpyHostToDevice);
```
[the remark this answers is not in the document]
r-abishek: It's defined in src/include/func_specific/spatter_mask.hpp

asalmanp (approved):

PR #88, opened 2022-01-19
Test suite fix for batchPD HIP

r-abishek:
Adds the hipDeviceSynchronize() after all hip function calls.

### DIFF
--- a/utilities/rpp-unittests/HIP_NEW/BatchPD_hip_pln1.cpp
+++ b/utilities/rpp-unittests/HIP_NEW/BatchPD_hip_pln1.cpp
@@ -765,10 +765,6 @@ int main(int argc, char **argv)
         else
             missingFuncFlag = 1;
 
-        hipDeviceSynchronize();
-
-        end = clock();
-
         break;
     }
     case 1:
@@ -799,10 +795,6 @@ int main(int argc, char **argv)
             missingFuncFlag = 1;
         else
             missingFuncFlag = 1;
-
-        hipDeviceSynchronize();
-
-        end = clock();
 
         break;
     }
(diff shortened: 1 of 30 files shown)


utilities/rpp-unittests/HIP_NEW/BatchPD_hip_pln1.cpp
```diff
@@ -765,6 +765,8 @@ int main(int argc, char **argv)
         else
             missingFuncFlag = 1;
 
+        hipDeviceSynchronize();
```
rrawther, line 768: Please move it outside the switch statement so you need to only make one line code change. Applies to all the files in this commit

rrawther (commented):
I think, you can move the hipDeviceSync.. () call outside the switch statement to have minimal code change.

utilities/rpp-unittests/HIP_NEW/BatchPD_hip_pln1.cpp
r-abishek: @rrawther The hipDeviceSynchronize() actually needs to be within the clock() timed section (before end = clock(); is called). Only then does the clock() wait for the GPU-offloads to finish entirely, and the end time can be recorded.

r-abishek: https://github.com/r-abishek/rpp/blob/ar/test_suite_mods/utilities/rpp-unittests/HIP_NEW/Tensor_hip_pkd3.cpp#L577-L598

rrawther: probably you can move both outside the switch statement in that case. It is better to keep common code all at one place

r-abishek: Ive now changed all instances in all test suite files to reflect this change. The end = clock() for HOST and the hipDeviceSynchronize() + end = clock() for HIP have been moved outside the switch case

rrawther (approved):

PR #89, opened 2022-01-31
Color Twist Tensor for HIP

r-abishek:
Adds color twist tensor implementation for HIP
Adds corresponding unit and performance tests

### DIFF
--- a/src/modules/hip/kernel/color_twist.hpp
+++ b/src/modules/hip/kernel/color_twist.hpp
@@ -150,110 +150,101 @@ __global__ void color_twist_pkd_tensor(T *srcPtr,
     rpp_hip_pack_float24_pln3_and_store24_pkd3(dstPtr, dstIdx, &pix_f24);
 }
 
-// template <typename T>
-// __global__ void color_twist_pln_tensor(T *srcPtr,
-//                                       uint3 srcStridesNCH,
-//                                       T *dstPtr,
-//                                       uint3 dstStridesNCH,
-//                                       RpptRGB *rgbTensor,
-//                                       float *alphaTensor,
-//                                       RpptROIPtr roiTensorPtrSrc)
-// {
-//     int id_x = (hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x) * 8;
-//     int id_y = hipBlockIdx_y * hipBlockDim_y + hipThreadIdx_y;
-//     int id_z = hipBlockIdx_z * hipBlockDim_z + hipThreadIdx_z;
+template <typename T>
+__global__ void color_twist_pln_tensor(T *srcPtr,
+                                      uint3 srcStridesNCH,
+                                      T *dstPtr,
+                                      uint3 dstStridesNCH,
+                                      float *brightnessTensor,
+                                      float *contrastTensor,
+                                      float *hueTensor,
+                                      float *saturationTensor,
+                                      RpptROIPtr roiTensorPtrSrc)
+{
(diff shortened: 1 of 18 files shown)


src/modules/hip/kernel/color_twist.hpp
```diff
@@ -0,0 +1,339 @@
+#include <hip/hip_runtime.h>
+#include "hip/rpp_hip_common.hpp"
+#include "cpu/rpp_cpu_common.hpp"
+
+__device__ void color_twist_1RGB_hip_compute(float *pixelR, float *pixelG, float *pixelB, float4 *colorTwistParams_f4)
```
rrawther, line 5: todo:: need to do this for 4 pixels to take advantage of vector math

src/modules/hip/kernel/color_twist.hpp
```diff
@@ -0,0 +1,339 @@
...
+{
+    float4 i8Offset_f4 = (float4) 128.0f;
+    float4 normalizer_f4 = (float4) ONE_OVER_255;
+    rpp_hip_math_add24_const(pix_f24, pix_f24, i8Offset_f4);
+    rpp_hip_math_multiply24_const(pix_f24, pix_f24, normalizer_f4);
+    colorTwistParams_f4->x = colorTwistParams_f4->x * 255.0f;
+    colorTwistParams_f4->z = (((int)colorTwistParams_f4->z) % 360) * SIX_OVER_360;
+    color_twist_8RGB_hip_compute(pix_f24, colorTwistParams_f4);
+    rpp_hip_pixel_check_0to255(pix_f24);
+    rpp_hip_math_subtract24_const(pix_f24, pix_f24, i8Offset_f4);
```
rrawther, line 118: what is i8Offset_f4 used for?

src/modules/rppt_tensor_augmentations.cpp
```diff
@@ -422,6 +422,76 @@ rppt_blend_host(RppPtr_t srcPtr1,
...
+                     RpptDescPtr dstDescPtr,
+                     Rpp32f *brightnessTensor,
+                     Rpp32f *contrastTensor,
+                     Rpp32f *hueTensor,
+                     Rpp32f *saturationTensor,
+                     RpptROIPtr roiTensorPtrSrc,
+                     RpptRoiType roiType,
+                     rppHandle_t rppHandle)
+{
+#ifdef HIP_COMPILE
```
rrawther, line 438: Is there a case for OPENCL as well?

src/modules/hip/kernel/color_twist.hpp
r-abishek: @rrawther Had a discussion with Aryan on this last week - Since the algorithm takes different directions for each different RGB pixel, a vector of 8 RGBs, might take n different paths and it wouldn't make sense to vectorize the algorithm without hip versions of vector logical operations or vector comparison operations like we do for vectors on the host with _mm_cmpge_ps() (greaterthan-equal) or _mm_or_ps(). To give a resultant mask for lanes that pass or fail the condition.

src/modules/hip/kernel/color_twist.hpp
r-abishek: Its just for the i8 range offset to -128 to +127.

src/modules/rppt_tensor_augmentations.cpp
r-abishek: No, not for any tensor based API.

asalmanp (approved):

PR #99, opened 2022-03-29
Resize Tensor on host - Bilinear Interpolation 

r-abishek:
This PR adds tensor resize bilinear interpolation for host and the respective unit/performance tests for resize bilinear interpolation and initial support for other interpolations.


utilities/rpp-performancetests/HOST_NEW/Tensor_host_pln1.cpp
```diff
@@ -25,6 +25,43 @@ typedef half Rpp16f;
 #define RPPMAX2(a,b) ((a > b) ? a : b)
 #define RPPMIN2(a,b) ((a < b) ? a : b)
 
+std::string get_interpolation_type(unsigned int val, RpptInterpolationType &interpolationType)
```
rrawther, line 28: consider putting this in a common header file instead of redefining in all .cpp files

utilities/rpp-unittests/HOST_NEW/Tensor_host_pkd3.cpp
```diff
@@ -24,6 +24,43 @@ typedef half Rpp16f;
 #define RPPMAX2(a,b) ((a > b) ? a : b)
 #define RPPMIN2(a,b) ((a < b) ? a : b)
 
+std::string get_interpolation_type(unsigned int val, RpptInterpolationType &interpolationType)
```
rrawther, line 27: If this is commonly used consider defining it in a common header file

rrawther (approved):
@paveltc to test

utilities/rpp-performancetests/HOST_NEW/Tensor_host_pln1.cpp
r-abishek: Will add this to the list of changes for our test suite changes. @sampath1117

paveltc:
@rrawther @r-abishek I tested this PR and found that everything seems to work as expected. Abishek, you mentioned that the images were resized to 100x180. I found that they were all resized to 203x74.

r-abishek:
@paveltc I just checked the test suite again. That sounds correct. Its actually resized to a factor of width/1.1 and height/3

PR #100, opened 2022-03-29
Restructure hip common

r-abishek:
@rrawther @asalmanp In this PR, all HIP functionalities were re-visited. New unions and structs have been used in hip to improve readability significantly with performance enhancements in certain variants.

@paveltc It affects code in all variants in Tensor HIP functionalities from ground up.

- Unions used are inspired from hip_vector_types.h and contain a data type ID to ensure the helpers are more intuitive
- The data type ID is:
    - ui1/ui2/ui3/ui4/ui6 for uint1/uint2/uint3/uint4/d_uint6
    - i1/i2/i3/i4/i6 for int1/int2/int3/int4/d_int6
    - f1/f2/f3/f4/f6/f8/f12/f16/f24 for float1/float2/float3/float4/d_float6/d_float8/d_float12/d_float16/d_float24
    - h1/h2/h3/h8/h24 for half/half2/hd_half8/d_half24 and others
- Every vector data type initialization in the hip common now has the specific format of variableNameInCamelCase_\<one of the dataTypeID from above\> for example srcPtr_f24 = float24 pointer or dstPtr_h8 = half8 pointer.
- It also adds 5%-40% performance improvements to certain filter type of functions - erode, dilate, box filter and for a few other functionalities.
- Adds few other minor changes for argument reductions, removal of old typedef struct usage in gridmask, warp_affine, filters.

### DIFF
--- a/src/include/hip/rpp_hip_common.hpp
+++ b/src/include/hip/rpp_hip_common.hpp
@@ -1149,7 +1149,7 @@ __device__ __forceinline__ void rpp_hip_load8_to_uchar8(uchar *srcPtr, uchar *sr
 
 __device__ __forceinline__ void rpp_hip_load8_to_uchar8(float *srcPtr, uchar *srcPtr_uc8)
 {
-    d_float8 src_f8;
+    d_float8 src_f8 = {0};
     *(d_float8_s *)&src_f8 = *(d_float8_s *)srcPtr;
 
     uint2 *srcPtr_ui2;
@@ -1200,7 +1200,7 @@ __device__ __forceinline__ void rpp_hip_load24_pkd3_to_uchar8_pln3(uchar *srcPtr
 
 __device__ __forceinline__ void rpp_hip_load24_pkd3_to_uchar8_pln3(float *srcPtr, uchar **srcPtrs_uc8)
 {
-    d_float24 src_f24;
+    d_float24 src_f24 = {0};
     *(d_float24_s *)&src_f24 = *(d_float24_s *)srcPtr;
 
     d_uint6 src_ui6;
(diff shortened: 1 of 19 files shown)


paveltc:
@r-abishek @asalmanp @rrawther I have run all of the relevant unit tests and checked them. Everything passes as expected.

rrawther (commented):
@paveltc there are lots of changes in this PR which affects load and store of many kernels. It is hard to verify all of that looking at the code. Is there a way to compare the output in binary from previous version so we can make it didn't introduce any bugs

paveltc:
@rrawther The first thing that comes to mind is to do a binary compare of the images from this PR with known good images that were gotten previously.
I did, however, visually verify all of the image results for this PR and didn't see anything that was wrong.

asalmanp (commented):
@r-abishek can you please check the remaining Codacy issues?

r-abishek:
@asalmanp I resolved most of them. There is only 1 pending:
- struct member 'd_schar8_s::sc1' is never used on L47 -> The custom type d_schar8_s is actually used in the next line L48

And secondly the only fix for the variable reassignment issue was to make an exception and change it to use the struct instead of the union in this commit -> https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/100/commits/0e4e22ab984cd7c968db7789ecf05b277f7ca0e1

asalmanp (approved):

src/include/hip/rpp_hip_common.hpp
```diff
@@ -9,207 +9,43 @@
...
-
-} d_schar12_as_schar3s;
-typedef struct d_schar24_as_schar3s
-{
-    d_schar12_as_schar3s x;
-    d_schar12_as_schar3s y;
-
-} d_schar24_as_schar3s;
+typedef struct { schar sc1[8];                                                                  }   d_schar8_s;
+typedef struct { d_schar8_s sc8[3];                                                             }   d_schar24_s;
```
rrawther, line 48: This definition can cause confusion for the compiler. It might interpret it as 2dim array. Can you define it as schar xx[24] instead

rrawther (changes requested):
@please fix the issue related to codacy warning

rrawther (approved):

PR #104, opened 2022-04-26
Contrast Tensor implementation on host and hip

r-abishek:
- Adds rpp support for contrast tensor on hip and host.
- Support for U8/F32/F16/I8, NCHW/NHWC/toggle is added.
- Adds corresponding unit tests and performance tests.


src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -2069,6 +2069,42 @@ inline RppStatus custom_convolve_image_host(T* srcPtr, RppiSize srcSize, U* dstP
 
 // Compute Functions for RPP Tensor API
 
+inline RppStatus compute_contrast_48_host(__m256 *p, __m256 *pContrastParams)
```
rrawther, line 2072: do we really need to return RppStatus in these inline functions? Compiler will just replace functions with the code. Please make those void

rrawther (changes requested):
@asalmanp to verify HIP kernels

src/include/cpu/rpp_cpu_common.hpp
r-abishek: To be changed in a separate PR as discussed - following the merge of #104 #105 and #106

rrawther (approved):
@paveltc : Please run all tests before merge

LakshmiKumar23:
@r-abishek @paveltc Works only with this docker
```
docker pull abishekr/rocm4.5.2_ubuntu18.04_rpp_mivisionx:v0.1
```
On my dev system, with rocm 5.1 and opencv 4.5, imwrite fails
On the rocm5.1 you provided, it fails with 
```
[3:00 PM] Ramasubramanian, Abishek
docker pull abishekr/rocm5.1_ubuntu18.04_rpp_mivisionx:v0.1.0
"hipErrorNoBinaryForGpu: Unable to find code object for all current devices!"
```

paveltc:
@r-abishek @rrawther I ran all of the tests and everything looks correct but there seems to be an extra test running for each collection of tests, contrast_u8_BatchPD. All the rest are labeled contrast_*_Tensor.
The BatchPD tests give different outputs than the Tensor ones, they all darken the image and seem to run some other augmentation.

r-abishek:
Thanks @paveltc. The unit tests still run both the old BatchPD and Tensor variants. Once a good volume of functions with the Tensor implementation is established, the BatchPD will be phased out. And yes, the contrast algorithm has changed to get better performance, so the old BatchPD algorithm is expected to be different from the new Tensor one.

And @LakshmiKumar23 I couldn't see any issue on the openCV side for either 4.5.5 or 3.4.0. I pulled the latest rocm 5.1.1 from DockerHub  and installed all dependencies on it for RPP. For OpenCV, I installed both versions in separate containers, (this time using the MIVisionX-setup.py instructions) and I could see the output dumps as expected.

paveltc:
@LakshmiKumar23 @rrawther @r-abishek I ran the unit tests on this PR using abishekr/rocm5.1.1_ubuntu18.04_opencv4.5.5_rpp_mivisionx:v0.1.0
Everything looks the same as with the previous run. As I understand it, the BatchPD is supposed to run for now.

PR #106, opened 2022-04-26
Salt and Pepper Noise Tensor implementation on host and hip

r-abishek:
- Adds rpp support for salt and pepper noise tensor on hip and host.
- Introduces generic 4 and 8 element seed-streams and support to running xorwow on SSE and AVX2 generating a vector of 8 random float numbers in parallel per CPU thread.
- Support for U8/F32/F16/I8, NCHW/NHWC/toggle is added.
- Adds corresponding unit tests and performance tests.

### DIFF
--- a/src/include/cpu/rpp_cpu_common.hpp
+++ b/src/include/cpu/rpp_cpu_common.hpp
@@ -32,10 +32,17 @@ typedef halfhpp Rpp16f;
 #define RPPISGREATER(pixel, value)      ((pixel > value) ? 1 : 0)
 #define RPPISLESSER(pixel, value)       ((pixel < value) ? 1 : 0)
 
+#if __AVX2__
+#define SIMD_FLOAT_VECTOR_LENGTH        8
+#else
+#define SIMD_FLOAT_VECTOR_LENGTH        4
+#endif
+
 static uint16_t wyhash16_x;
 
 alignas(64) const Rpp32f sch_mat[16] = {0.701f, -0.299f, -0.300f, 0.0f, -0.587f, 0.413f, -0.588f, 0.0f, -0.114f, -0.114f, 0.886f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f};
 alignas(64) const Rpp32f ssh_mat[16] = {0.168f, -0.328f, 1.250f, 0.0f, 0.330f, 0.035f, -1.050f, 0.0f, -0.497f, 0.292f, -0.203f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f};
+alignas(64) const Rpp32u multiseedStreamOffset[8] = {1436021U, 2316707U, 4377697U, 1648039U, 1194659U, 2224457U, 3005987U, 2790331U};
 
 inline uint32_t hash16(uint32_t input, uint32_t key) {
   uint32_t hash = input * key;
@@ -75,36 +82,13 @@ inline int fastrand()
     return (g_seed>>16)&0x7FFF;
 }
 
-inline void rpp_host_rng_xorwow_f32_initialize_8seed_stream(RpptXorwowState *xorwowInitialState, Rpp32u seed)
+template<Rpp32s STREAM_SIZE>
+inline void rpp_host_rng_xorwow_f32_initialize_multiseed_stream(RpptXorwowState *xorwowInitialState, Rpp32u seed)
 {
-    Rpp32u xorwowSeedStream[8];
-    xorwowSeedStream[0] = seed + 1436017U;
-    xorwowSeedStream[1] = seed + 2316719U;
-    xorwowSeedStream[2] = seed + 4377713U;
-    xorwowSeedStream[3] = seed + 1648031U;
-    xorwowSeedStream[4] = seed + 1194655U;
-    xorwowSeedStream[5] = seed + 2224453U;
-    xorwowSeedStream[6] = seed + 3005991U;
-    xorwowSeedStream[7] = seed + 2790345U;
-    for (int i = 0; i < 8; i++)
-    {
-        xorwowInitialState[i].x[0] = 123456789U + xorwowSeedStream[i];
-        xorwowInitialState[i].x[1] = 362436069U + xorwowSeedStream[i];
-        xorwowInitialState[i].x[2] = 521288629U + xorwowSeedStream[i];
-        xorwowInitialState[i].x[3] = 88675123U + xorwowSeedStream[i];
-        xorwowInitialState[i].x[4] = 5783321U + xorwowSeedStream[i];
-        xorwowInitialState[i].counter = 6615241U + xorwowSeedStream[i];
-    }
-}
-
-inline void rpp_host_rng_xorwow_f32_initialize_4seed_stream(RpptXorwowState *xorwowInitialState, Rpp32u seed)
-{
-    Rpp32u xorwowSeedStream[4];
-    xorwowSeedStream[0] = seed + 1436017U;
-    xorwowSeedStream[1] = seed + 2316719U;
-    xorwowSeedStream[2] = seed + 4377713U;
-    xorwowSeedStream[3] = seed + 1648031U;
-    for (int i = 0; i < 4; i++)
+    Rpp32u xorwowSeedStream[STREAM_SIZE];
+    for (int i = 0; i < STREAM_SIZE; i++)
+        xorwowSeedStream[i] = seed + multiseedStreamOffset[i];
+    for (int i = 0; i < STREAM_SIZE; i++)
     {
         xorwowInitialState[i].x[0] = 123456789U + xorwowSeedStream[i];
         xorwowInitialState[i].x[1] = 362436069U + xorwowSeedStream[i];
--- a/src/include/cpu/rpp_cpu_common.hpp
+++ b/src/include/cpu/rpp_cpu_common.hpp
@@ -42,7 +42,7 @@ static uint16_t wyhash16_x;
 
 alignas(64) const Rpp32f sch_mat[16] = {0.701f, -0.299f, -0.300f, 0.0f, -0.587f, 0.413f, -0.588f, 0.0f, -0.114f, -0.114f, 0.886f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f};
 alignas(64) const Rpp32f ssh_mat[16] = {0.168f, -0.328f, 1.250f, 0.0f, 0.330f, 0.035f, -1.050f, 0.0f, -0.497f, 0.292f, -0.203f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f};
-alignas(64) const Rpp32u multiseedStreamOffset[8] = {1436021U, 2316707U, 4377697U, 1648039U, 1194659U, 2224457U, 3005987U, 2790331U};
+alignas(64) const Rpp32u multiseedStreamOffset[8] = {0x15E975, 0x2359A3, 0x42CC61, 0x1925A7, 0x123AA3, 0x21F149, 0x2DDE23, 0x2A93BB};    // Prime numbers for multiseed stream initialization
 
 inline uint32_t hash16(uint32_t input, uint32_t key) {
   uint32_t hash = input * key;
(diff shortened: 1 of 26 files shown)


src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -75,6 +75,106 @@ inline int fastrand()
     return (g_seed>>16)&0x7FFF;
 }
 
+inline void rpp_host_rng_xorwow_f32_initialize_8seed_stream(RpptXorwowState *xorwowInitialState, Rpp32u seed)
```
rrawther, line 78: it is better to define it as template function with #seeds as template parameter for 4 seed and 8 seed cases

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -75,6 +75,106 @@ inline int fastrand()
...
+        xorwowInitialState[i].counter = 6615241U + xorwowSeedStream[i];
+    }
+}
+
+inline __m256 rpp_host_rng_xorwow_8_f32_avx(__m256i *pxXorwowStateXParam, __m256i *pxXorwowStateCounterParam)
+{
+    __m256i px362437 = _mm256_set1_epi32(362437);
+    __m256i pxFFFFFFFF = _mm256_set1_epi32(0xFFFFFFFF);
+    __m256i px7FFFFF = _mm256_set1_epi32(0x7FFFFF);
+    __m256i pxExponentFloat = _mm256_set1_epi32(0b111111100000000000000000000000);
```
rrawther, line 123: please use hex value (0x) for set1_epi32

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -75,6 +75,106 @@ inline int fastrand()
...
+    __m256i pxT = pxXorwowStateXParam[4];
+    __m256i pxS = pxXorwowStateXParam[0];
+    pxXorwowStateXParam[4] = pxXorwowStateXParam[3];
```
rrawther, line 126: are all these values supposed to be the same. There is lot of ambiguity here

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -75,6 +75,106 @@ inline int fastrand()
...
+    pxT = _mm_xor_si128(pxT, _mm_xor_si128(pxS, _mm_slli_epi32(pxS, 4)));
+    pxXorwowStateXParam[0] = pxT;
+    *pxXorwowStateCounterParam = _mm_and_si128(_mm_add_epi32(*pxXorwowStateCounterParam, px362437), pxFFFFFFFF);
+    pxS = _mm_or_si128(pxExponentFloat, _mm_and_si128(_mm_add_epi32(pxT, *pxXorwowStateCounterParam), px7FFFFF));
+    return _mm_sub_ps(*(__m128 *)&pxS, xmm_p1);
+}
+
+inline float rpp_host_rng_xorwow_f32(RpptXorwowState *xorwowState)
+{
+    uint t  = xorwowState->x[4];
```
rrawther, line 162: I would like to review this separate. It needs rework

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1410,6 +1410,38 @@ __device__ __forceinline__ void rpp_hip_math_subtract24_const(d_float24 *src_f24
     dst_f24->f4[5] = src_f24->f4[5] - subtrahend_f4;
 }
 
+// /******************** DEVICE RANDOMIZATION HELPER FUNCTIONS ********************/
+
+__device__ __forceinline__ float rpp_hip_rng_xorwow_f32(RpptXorwowState *xorwowState)
+{
+    uint t  = xorwowState->x[4];
```
rrawther, line 1417: need to rewrite this function. Not sure why x[1]..x[4] are all the same

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1410,6 +1410,38 @@ __device__ __forceinline__ void rpp_hip_math_subtract24_const(d_float24 *src_f24
...
+    xorwowState->x[0] = t;
+    xorwowState->counter = (xorwowState->counter + 362437) & 0xFFFFFFFF;
+    uint out = (0b111111100000000000000000000000 | ((t + xorwowState->counter) & 0x7fffff));
+    float outFloat = *(float *)&out;
+    return  outFloat - 1;
+}
+
+__device__ __forceinline__ void rpp_hip_rng_8_xorwow_f32(RpptXorwowState *xorwowState, d_float8 *randomNumbersPtr_f8)
+{
+    randomNumbersPtr_f8->f1[0] = rpp_hip_rng_xorwow_f32(xorwowState);
```
rrawther, line 1435: are this used inside a kernel? Why is this not done in multiple threads?

src/modules/cpu/kernel/noise_salt_and_pepper.hpp
```diff
@@ -0,0 +1,1646 @@
...
+                }
+                for (; vectorLoopCount < bufferLength; vectorLoopCount++)
+                {
+                    Rpp16f src[3] = {*srcPtrTempR, *srcPtrTempG, *srcPtrTempB};
+                    float randomNumberFloat = rpp_host_rng_xorwow_f32(&xorwowState);
+                    if (randomNumberFloat > noiseProbability)
+                        memcpy(dstPtrTemp, src, 6);
+                    else
+                        if(randomNumberFloat <= saltProbability)
+                            std::fill_n(dstPtrTemp, 3, salt);
```
rrawther, line 1017: please use memset instead

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pkd3.cpp
```diff
@@ -58,6 +58,17 @@ std::string get_interpolation_type(unsigned int val, RpptInterpolationType &inte
     }
 }
 
+std::string get_noise_type(unsigned int val)
```
rrawther, line 61: please use lut instead of conditional branch

src/include/cpu/rpp_cpu_common.hpp
r-abishek: @rrawther Actually all these values aren't the same. According to the algorithm, at every run of the xorwow(), it needs 5 state values x[0-4] and a sixth param called counter. To compute the next state of the rng, the new x[1-4] are replaced with the old x[0-3] and the new x[0] is computed at the end.
And the counter has a separate updation.

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pkd3.cpp
```diff
@@ -759,6 +781,88 @@ int main(int argc, char **argv)
...
+                    else if (ip_bitDepth == 5)
+                        rppt_salt_and_pepper_noise_gpu(d_inputi8, srcDescPtr, d_outputi8, dstDescPtr, noiseProbabilityTensor, saltProbabilityTensor, saltValueTensor, pepperValueTensor, seed, d_roiTensorPtrSrc, roiTypeSrc, handle);
+                    else if (ip_bitDepth == 6)
+                        missingFuncFlag = 1;
+                    else
+                        missingFuncFlag = 1;
+
+                    break;
+                }
+                case 1:
```
rrawther, line 847: please collapse all same definition switch statements into default

rrawther (changes requested):
we need to combine most test case variations into single file. Every PR has so many test file changes.

src/modules/cpu/kernel/noise_salt_and_pepper.hpp
r-abishek: I did have meemset here before (like the U8 and I8 cases). But since this is the F32/F16 variant for noise, we need to set 0-1 float values. 
However the second argument for memset is integer and the float values were being floored to 0. So needed another equivalent of memset that works with float.

src/include/cpu/rpp_cpu_common.hpp

src/include/cpu/rpp_cpu_common.hpp

src/include/cpu/rpp_cpu_common.hpp
r-abishek: x-params reordered in the xorwow state

src/include/hip/rpp_hip_common.hpp
r-abishek: same as https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/106#discussion_r862036992

src/include/hip/rpp_hip_common.hpp
r-abishek: combined rgb computations for 3 channel variants

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pkd3.cpp
r-abishek: Will add this to the test suite PR

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pkd3.cpp
r-abishek: Will add this to the test suite PR

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -75,6 +84,106 @@ inline int fastrand()
...
+
+    // Update all 6 xorwow state params
+    pxXorwowStateXParam[0] = pxXorwowStateXParam[1];                                                                        // xorwowState->x[0] = xorwowState->x[1];
+    pxXorwowStateXParam[1] = pxXorwowStateXParam[2];                                                                        // xorwowState->x[1] = xorwowState->x[2];
+    pxXorwowStateXParam[2] = pxXorwowStateXParam[3];                                                                        // xorwowState->x[2] = xorwowState->x[3];
+    pxXorwowStateXParam[3] = pxXorwowStateXParam[4];                                                                        // xorwowState->x[3] = xorwowState->x[4];
+    pxXorwowStateXParam[4] = pxT;                                                                                           // xorwowState->x[4] = t;
+    *pxXorwowStateCounterParam = _mm_and_si128(_mm_add_epi32(*pxXorwowStateCounterParam, px362437), pxFFFFFFFF);            // xorwowState->counter = (xorwowState->counter + 362437) & 0xFFFFFFFF;
```
rrawther, line 157: better to represent hex number as 0x..

src/include/hip/rpp_hip_common.hpp
```diff
@@ -80,11 +80,14 @@ struct RPPTensorFunctionMetaData
...
+#define LOCAL_THREADS_X                 16
+#define LOCAL_THREADS_Y                 16
+#define LOCAL_THREADS_Z                 1
+#define ONE_OVER_255                    0.00392157f
+#define SIX_OVER_360                    0.01666667f
+#define XORWOW_COUNTER_INC              0x587C5     // Hex 0x587C5 = Dec 362437U - xorwow counter increment
+#define XORWOW_EXPONENT_MASK            0x3F800000  // Hex 0x3F800000 = Bin 0b111111100000000000000000000000 - 23 bits of mantissa set to 0, 01111111 for the exponent, 0 for the sign bit
+
 #define BYTE_TO_BINARY_PATTERN "%c%c%c%c%c%c%c%c"
 #define BYTE_TO_BINARY(byte)  \
```
rrawther, line 92: what is the use of this macro?

rrawther (changes requested):
@paveltc to test this

src/include/cpu/rpp_cpu_common.hpp
r-abishek: done

src/include/hip/rpp_hip_common.hpp
r-abishek: That one had been added for hip debugging purposes. It helps display U8/F32 pixels in binary. (Could be removed now, or later once more tensor hip kernels are added)

paveltc:
@rrawther @r-abishek All of the unit tests pass for this PR. This PR also runs the u8_BatchPD test for some reason. It's possible there are some differences between the results of the BatchPD and the Tensor tests.

r-abishek:
@rrawther @paveltc Thanks Pavel, this Salt and Pepper Noise PR also has fundamental changes in the algorithm for better performance, and it's expected that the Tensor outputs shouldn't match the older BatchPD outputs.

paveltc:
@LakshmiKumar23 @rrawther @r-abishek Ran the unit tests on this with the abishekr/rocm5.1.1_ubuntu18.04_opencv4.5.5_rpp_mivisionx:v0.1.0. Everything looks fine.

rrawther (approved):
@paveltc to confirm it is passing all the tests

PR #109, opened 2022-05-24
Change return type of inline helpers from RppStatus to void

r-abishek:
@rrawther This PR just changes the return type of inline functions from RppStatus RPP_SUCCESS to void as discussed.

1. Change return type of inline functions from RppStatus to void
2. Remove the RPP_SUCCESS return statement for all these inline void functions

### DIFF
--- a/src/include/cl/rpp_cl_common.hpp
+++ b/src/include/cl/rpp_cl_common.hpp
@@ -54,7 +54,7 @@ inline int getplnpkdind(RppiChnFormat &format)
     return format == RPPI_CHN_PLANAR ? 1 : 3;
 }
 
-inline RppStatus generate_gaussian_kernel_gpu(Rpp32f stdDev, Rpp32f *kernel, Rpp32u kernelSize)
+inline void generate_gaussian_kernel_gpu(Rpp32f stdDev, Rpp32f *kernel, Rpp32u kernelSize)
 {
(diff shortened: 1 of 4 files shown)


src/include/cl/rpp_cl_common.hpp
```diff
@@ -74,7 +74,5 @@ inline RppStatus generate_gaussian_kernel_gpu(Rpp32f stdDev, Rpp32f *kernel, Rpp
     {
         kernel[i] /= sum;
```
rrawther, line 75: this can be done only if sum is guaranteed to be nonzero

rrawther (commented):
@paveltc  to run tests
@r-abishek check possibility of division by zero

r-abishek:
@paveltc This PR doesn't add any functionality, and just changes some return types, so just verifying a successful build for OpenCL and HIP are relevant.
@rrawther The sum is dependent on a non-zero factor and exp(x). And exp(x) cannot be zero, so the sum won't be zero.

paveltc:
@rrawther @r-abishek 
I have built this PR with both HIP and OpenCL backends. Both build fine.
I have also built MIVisionX afterwards, it also builds fine for both but I did not have the dependencies installed for OpenCL so parts of MIVisionX were left out. As far as I know, the parts that get left out don't interact with RPP.

rrawther (approved):

PR #120, opened 2022-08-02
Shot and Gaussian Noise

r-abishek:
- Introduces shot and gaussian noise functionalities for HIP and HOST
- Adds relevant unit and perf tests


kiritigowda:
@r-abishek -- can you take a look at these - https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/120/checks?check_run_id=7624245942

r-abishek:
> @r-abishek -- can you take a look at these - https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/120/checks?check_run_id=7624245942

@kiritigowda I had looked at them after issuing the PR, actually these casts from float* to int* and back to float* are essential to the working of the Fast Inverse Square Root algorithm. Wasn't sure if there could be a workaround for them:
https://www.wikiwand.com/en/Fast_inverse_square_root
gy4VwrfW2MDeGW0-49BeoRtIurXDPjH-5xzQneMrUNwDaENjXEVEZPMQBy1QFpoOv7VQ12yjP1wk08IP_uu0Qx5rTLqfgNvTSktvVt7ozz6hqUUe402sYnRv3hIN

r-abishek:
@rrawther This is the block of code computing the fast inverse square root.
https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/120/files#diff-9b79ec15bdd883616f45c7f27b3577960efa34d4e4dec79dc3236722f1b33a9cR155-R190

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1424,9 +1470,109 @@ __device__ __forceinline__ void rpp_hip_math_subtract24_const(d_float24 *src_f24
...
+    dst_f24->f4[2] = src_f24->f4[2] * multiplier_f4;
+    dst_f24->f4[3] = src_f24->f4[3] * multiplier_f4;
+    dst_f24->f4[4] = src_f24->f4[4] * multiplier_f4;
+    dst_f24->f4[5] = src_f24->f4[5] * multiplier_f4;
+}
+
+__device__ __forceinline__ float rpp_hip_math_inverse_sqrt1(float x)
+{
+    float xHalf = 0.5f * x;
+    int i = *(int*)&x;                              // float bits in int
```
asalmanp, line 1516: please use the builtin hip functions to convert float to int (__float2int_rn)

asalmanp (changes requested):

src/include/hip/rpp_hip_common.hpp
r-abishek: Actually as in the above comments, this inverse_sqrt algorithm needs the same float bits (same order of sign-exponent-mantissa in IEEE format) but to be interpreted as int for the right shifting. But the __float2int_rn() may end up converting the value itself to integer.

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -130,6 +152,52 @@ struct RPPTensorFunctionMetaData
 };
 #endif // GPU_SUPPORT
 
+inline float rpp_host_math_inverse_sqrt_1(float x)
```
rrawther, line 155: please add a comment with the link to the the paper taking this algorithm

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -151,83 +219,218 @@ inline void rpp_host_rng_xorwow_f32_initialize_multiseed_stream(RpptXorwowState
...
+inline Rpp32f rpp_host_rng_xorwow_f32(T *xorwowState)
+{
+    // Update xorwow state
+    rpp_host_rng_xorwow_state_update(xorwowState);
 
     // Create float representation and return 0 <= outFloat < 1
-    uint out = (XORWOW_EXPONENT_MASK | ((t + xorwowState->counter) & 0x7FFFFF));        // bitmask 23 mantissa bits, OR with exponent
-    float outFloat = *(float *)&out;                                                    // reinterpret out as float
-    return  outFloat - 1;                                                               // return 0 <= outFloat < 1
+    xorwowState->counter &= 0xFFFFFFFF;                                                             // set new state param counter
```
rrawther, line 377: why is this required. Looks like it doesn't change the value

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -151,83 +219,218 @@ inline void rpp_host_rng_xorwow_f32_initialize_multiseed_stream(RpptXorwowState
...
+    pV = _mm_fmadd_ps(pV, xmm_p2Pow32InvMul2Pi, xmm_p2Pow32InvMul2PiDiv2);                      // v = v * RPP_2POW32_INV_MUL_2PI + RPP_2POW32_INV_MUL_2PI_DIV_2;
+    pS = _mm_div_ps(xmm_p1, rpp_host_math_inverse_sqrt_4_sse(_mm_mul_ps(xmm_pm2, log_ps(pU)))); // s = 1 / rpp_host_math_inverse_sqrt_1(-2.0f * std::log(u));
+    sincos_ps(pV, &pU, &pV);                                                                    // std::sin(v) and std::cos(v) computation
+    pRngVals[0] = _mm_mul_ps(pU, pS);                                                           // u = std::sin(v) * s;
+    pRngVals[1] = _mm_mul_ps(pV, pS);                                                           // v = std::cos(v) * s;
+}
+
+inline float rpp_host_rng_1_gaussian_f32(RpptXorwowStateBoxMuller *xorwowState)
+{
+    if(xorwowState->boxMullerFlag == 0)
```
rrawther, line 419: if (!xorwowState->boxMullerFlag) is better here

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -151,83 +219,218 @@ inline void rpp_host_rng_xorwow_f32_initialize_multiseed_stream(RpptXorwowState
...
+    {
+        Rpp32f u, v, s;
+        u = (Rpp32f)rpp_host_rng_xorwow_u32(xorwowState) * RPP_2POW32_INV + RPP_2POW32_INV_DIV_2;
+        v = (Rpp32f)rpp_host_rng_xorwow_u32(xorwowState) * RPP_2POW32_INV_MUL_2PI + RPP_2POW32_INV_MUL_2PI_DIV_2;
+        s = 1 / rpp_host_math_inverse_sqrt_1(-2.0f * std::log(u));
```
rrawther, line 424: this is an inverse operation and will affect performance. Explore ways to avoid this

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -151,83 +219,218 @@ inline void rpp_host_rng_xorwow_f32_initialize_multiseed_stream(RpptXorwowState
...
+    pRngVals[0] = _mm_mul_ps(pU, pS);                                                           // u = std::sin(v) * s;
+    pRngVals[1] = _mm_mul_ps(pV, pS);                                                           // v = std::cos(v) * s;
+}
+
+inline float rpp_host_rng_1_gaussian_f32(RpptXorwowStateBoxMuller *xorwowState)
+{
+    if(xorwowState->boxMullerFlag == 0)
+    {
+        Rpp32f u, v, s;
+        u = (Rpp32f)rpp_host_rng_xorwow_u32(xorwowState) * RPP_2POW32_INV + RPP_2POW32_INV_DIV_2;
```
rrawther, line 422: do rpp_host_rng_xorwow_u32(xorwowState) only once and reuse for both statements

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -2953,6 +3156,183 @@ inline void compute_salt_and_pepper_noise_48_host(__m128 *p, __m128i *pxXorwowSt
...
+    __m128 pFactValue = fast_exp_sse(_mm_mul_ps(*p, *pShotNoiseFactorInv));                                                                             // Rpp32f factValue = expf(lambda);
+    __m128 pFactValueMask = xmm_p1;                                                                                                                     // Set mask for pFactValue computation to 1
+    do
+    {
+        pShotNoiseValue = _mm_or_ps(_mm_andnot_ps(pFactValueMask, pShotNoiseValue), _mm_and_ps(pFactValueMask, _mm_add_ps(pShotNoiseValue, xmm_p1)));   // shotNoiseValue++;
+        pFactValue = _mm_mul_ps(pFactValue, rpp_host_rng_xorwow_4_f32_sse(pxXorwowStateX, pxXorwowStateCounter));                                       // factValue *= rpp_host_rng_xorwow_f32(xorwowStatePtr);
+        pFactValueMask = _mm_cmpgt_ps(pFactValue, xmm_p1);                                                                                              // compute new pFactValueMask for loop exit condition
+    } while (_mm_movemask_epi8(_mm_cvtps_epi32(pFactValueMask)) != 0);                                                                                  // while (factValue > 1.0f);
```
rrawther, line 3185: not sure what the condition is doing for while()?

src/include/func_specific/rng_seed_stream.hpp
```diff
@@ -5,7 +5,7 @@
...
... 155259, 764716, 235733, 481317, 306885, 732218, 115590, 917449, 723326, 324296, 878162, 805300, 865740, 830788, 319344, 297076, 630370, 386384, 780253, 100584, 731605, 134795, 820104, 121115, 514356, 123256, 400428, 446013, 151280, 916043, 345287, 379787, 129565, 947156, 974294, 707583, 910553, 130917, 981660, 146049, 793212, 651934, 537031, 168055, 823490, 806687, 503217, 471696, 593786, 102956, 558731, 276764, 643359, 289622, 957489, 619223, 651677, 191795, 283771, 155046, 186852, 310593, 121445, 343704, 153587, 293532, 380375, 321084, 595166, 161268, 944827, 542115, 316626, 428468, 130162, 684777, 542142, 998308, 150446, 575299, 340353, 977809, 837318, 734692, 613815, 214647, 462703, 226586, 863995, 563267, 734383, 573703, 891686, 973411, 879183, 590344, 150313, 747737, 968821, 200808, 589146, 231348, 533529, 575375, 256785, 403840, 775818, 939712, 618134, 539406, 267952, 176201, 267510, 414783, 177672, 152609, 702173, 198224, 547158, 469752, 507664, 911999, 449574, 894131, 720221, 446580, 120735, 231717, 980608, 850706, 132276, 944840, 590995, 582994, 761474, 957166, 239989, 309587, 190973, 836071, 381133, 124200, 711759, 334858, 495950, 590018, 362332, 615438, 179266, 9505798
```
rrawther, line 8: avoid using so wide statements. Please split it into multiple lines for readability

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1412,6 +1440,24 @@ __device__ __forceinline__ void rpp_hip_math_add24_const(d_float24 *src_f24, d_f
...
+    dstPtr_f16->f4[1] = src1Ptr_f16->f4[1] - src2Ptr_f16->f4[1];
+    dstPtr_f16->f4[2] = src1Ptr_f16->f4[2] - src2Ptr_f16->f4[2];
+    dstPtr_f16->f4[3] = src1Ptr_f16->f4[3] - src2Ptr_f16->f4[3];
+}
+
+// d_float8 subtract with constant
+
+__device__ __forceinline__ void rpp_hip_math_subtract8_const(d_float8 *src_f8, d_float8 *dst_f8, float4 subtrahend_f4)
+{
+    dst_f8->f4[0] = src_f8->f4[0] + subtrahend_f4;
```
rrawther, line 1457: Is this addition or subtraction?

src/include/hip/rpp_hip_common.hpp
rrawther: @asalmanp: It is not really converting from float to int. It is just representing the exact float number as int.
@r-abishek Please add link to the paper for the algorithm

src/modules/cpu/kernel/noise_gaussian.hpp
```diff
@@ -0,0 +1,1442 @@
...
+        Rpp8u *srcPtrChannel, *dstPtrChannel;
+        srcPtrChannel = srcPtrImage + (roi.xywhROI.xy.y * srcDescPtr->strides.hStride) + (roi.xywhROI.xy.x * layoutParams.bufferMultiplier);
+        dstPtrChannel = dstPtrImage;
+
+        Rpp32u alignedLength = (bufferLength / 48) * 48;
+        Rpp32u vectorIncrement = 48;
+        Rpp32u vectorIncrementPerChannel = 16;
+
+        RpptXorwowStateBoxMuller xorwowState;
+        xorwowState.x[0] = xorwowInitialStatePtr[0].x[0] + offset;
```
rrawther, line 45: tbd:: keep this in vector of 8 and use AVX to update

src/modules/cpu/kernel/noise_shot.hpp
```diff
@@ -0,0 +1,1666 @@
...
+
+        Rpp8u *srcPtrChannel, *dstPtrChannel;
+        srcPtrChannel = srcPtrImage + (roi.xywhROI.xy.y * srcDescPtr->strides.hStride) + (roi.xywhROI.xy.x * layoutParams.bufferMultiplier);
+        dstPtrChannel = dstPtrImage;
+
+        Rpp32u alignedLength = (bufferLength / 48) * 48;
+        Rpp32u vectorIncrement = 48;
+        Rpp32u vectorIncrementPerChannel = 16;
+
+        RpptXorwowState xorwowState;
```
rrawther, line 43: line 44-48 can be done in one vector instruction

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -1439,6 +1669,128 @@ static inline __m128 atan2_ps( __m128 y, __m128 x )
...
+    y = _mm256_sub_ps(y, tmp);
+    tmp = _mm256_mul_ps(e, *(__m256 *)&avx_cephesLogQ2);
+    x = _mm256_add_ps(x, y);
+    x = _mm256_add_ps(x, tmp);
+    x = _mm256_or_ps(x, invalid_mask); // negative arg will be NAN
+
+    return x;
+}
+
+static inline __m128 log_ps(__m128 x)
```
rrawther, line 1733: Is there is reference link to the algorithm here? Please include. Otherwise insert comment on what this is doing

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -151,83 +219,218 @@ inline void rpp_host_rng_xorwow_f32_initialize_multiseed_stream(RpptXorwowState
...
+    Rpp32u xorwowSeedStream[STREAM_SIZE];
+
+    // Loop to initialize seed stream of size STREAM_SIZE based on user seed and offset
+    for (int i = 0; i < STREAM_SIZE; i++)
+        xorwowSeedStream[i] = seed + multiseedStreamOffset[i];
+
+    // Loop to initialize STREAM_SIZE xorwow initial states for multi-stream random number generation
+    for (int i = 0; i < STREAM_SIZE; i++)
+    {
+        xorwowInitialState[i].x[0] = 0x75BCD15 + xorwowSeedStream[i];      // state param x[0] offset 123456789U
```
rrawther, line 234: please add comments on what these magic numbers mean

src/modules/cpu/kernel/noise_shot.hpp
```diff
@@ -0,0 +1,1666 @@
...
+                dstPtrTempG = dstPtrRowG;
+                dstPtrTempB = dstPtrRowB;
+
+                int vectorLoopCount = 0;
+                for (; vectorLoopCount < alignedLength; vectorLoopCount += vectorIncrement)
+                {
+#if __AVX2__
+                    __m256 p[6];
+                    rpp_simd_load(rpp_load48_u8pkd3_to_f32pln3_avx, srcPtrTemp, p);    // simd loads
+                    if (shotNoiseFactor != 0.0f)
```
rrawther, line 100: if (shotNoiseFactor) would suffice here and perform better

src/modules/cpu/kernel/noise_shot.hpp
```diff
@@ -0,0 +1,1666 @@
+#include "rppdefs.h"
+#include "rpp_cpu_simd.hpp"
+#include "rpp_cpu_common.hpp"
+
+RppStatus shot_noise_u8_u8_host_tensor(Rpp8u *srcPtr,
```
rrawther, line 5: This file is too large with very similar code in many functions. We need better code so it won't replicate code in all the functions

src/modules/cpu/kernel/noise_gaussian.hpp
```diff
@@ -0,0 +1,1442 @@
+#include "rppdefs.h"
+#include "rpp_cpu_simd.hpp"
+#include "rpp_cpu_common.hpp"
+
+RppStatus gaussian_noise_u8_u8_host_tensor(Rpp8u *srcPtr,
```
rrawther, line 5: too huge file with similar code repeated in many functions. We need better ways to condense it.

rrawther (commented):
@r-abishek : This PR is very huge and hard to review
@asalmanp to review hip kernels
@r-abishek : Need to condense all test suits into single one as mentioned before. Every new functionality has so many test suit files to change since we have many files. Most of them are similar and can be combined with appropriate arguments

src/include/cpu/rpp_cpu_common.hpp
r-abishek: Added citation - Lomont, C., 2003. FAST INVERSE SQUARE ROOT. [online] lomont.org. Available at: <http://www.lomont.org/papers/2003/InvSqrt.pdf>

src/include/cpu/rpp_cpu_common.hpp
r-abishek: It was in the reference algorithm too, but yeah its just masking all 32 bits. Removed now.

src/include/cpu/rpp_cpu_common.hpp

src/include/cpu/rpp_cpu_common.hpp
r-abishek: Removed this 👍. Explored a couple options, looks like it might be better to retain inverse sqrt for hip side and just use SSE/AVX2 sqrt intrinsic for host based on the performance numbers.

src/include/cpu/rpp_cpu_common.hpp
r-abishek: This may not be possible since we need two runs of the randomization as per the cuda approach. The rng's state has changed after the first run, and its used to generate the next number

src/include/cpu/rpp_cpu_common.hpp
r-abishek: This SSE while loop is just replicating the equivalent serial code which is while (factValue > 1.0f); The loop needs to run until all values in the vector satisfy that.

src/include/func_specific/rng_seed_stream.hpp
r-abishek: Done

src/include/hip/rpp_hip_common.hpp
r-abishek: Thanks, bug fixed

src/modules/cpu/kernel/noise_gaussian.hpp
r-abishek: Ran a few experiments on this, turns out it doesn't make a difference if all the 5-6 params in the xorwow state are updated with the offset. We just need one param to be updated and that should suffice to achieve the desired randomization. So, currently updating just x[0] with offset.

src/modules/cpu/kernel/noise_shot.hpp
r-abishek: Changed these lines, as well as a couple of the following lines to inline functions. 👍 Those inline helpers are now called by 12 functions (shotNoise-u8/f32/f16/i8, gaussianNoise-u8/f32/f16/i8 and saltAndPepperNoise-u8/f32/f16/i8)

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: SSE reference added - https://github.com/RJVB/sse_mathfun/blob/master/sse_mathfun.h

src/include/cpu/rpp_cpu_common.hpp
r-abishek: Original citation for magic numbers added - Marsaglia, G. (2003). Xorshift RNGs. Journal of Statistical Software, 8(14), 1–6. https://doi.org/10.18637/jss.v008.i14

src/modules/cpu/kernel/noise_shot.hpp
r-abishek: Done

src/modules/cpu/kernel/noise_shot.hpp
r-abishek: A couple hundred lines were reduced in those files through this change - https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/120#discussion_r947390200

src/modules/cpu/kernel/noise_gaussian.hpp
r-abishek: Same as prev. Similar lines also reduced in salt_and_pepper noise.

r-abishek:
> @r-abishek : This PR is very huge and hard to review @asalmanp to review hip kernels @r-abishek : Need to condense all test suits into single one as mentioned before. Every new functionality has so many test suit files to change since we have many files. Most of them are similar and can be combined with appropriate arguments

@rrawther I understand, this specific PR got a little bigger than usual since the two noise functionalities were related in their working, and had common helpers, so I ended up putting them in one PR so its easier to verify that the helper functions I write work correctly for both the functions. 
Regarding the test suite, our intention is to change the whole testsuite so it takes user params from a file like GDF, with minimal code change to add new functionalities. But in the mean time I didn't want to stop the flow for regular PRs on adding new functionalities. As soon as the test suite has a readiness level taking care of all functionalities, that'll be a separate PR and we won't need so many test suite changes.

paveltc:
@r-abishek @rrawther @asalmanp @kiritigowda This PR passes all relevant unit tests.

src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -130,6 +152,55 @@ struct RPPTensorFunctionMetaData
...
+inline __m128 rpp_host_math_inverse_sqrt_4_sse(__m128 p)
+{
+    __m128 pHalfNeg;
+    __m128i pxI;
+    pHalfNeg = _mm_mul_ps(_ps_n0p5, p);                                         // float xHalfNeg = -0.5f * x;
+    pxI = *(__m128i *)&p;                                                       // int i = *(int*)&x;
+    pxI = _mm_sub_epi32(xmm_newtonMethodInitialGuess, _mm_srli_epi32(pxI, 1));  // i = NEWTON_METHOD_INITIAL_GUESS - (i >> 1);
+    p = *(__m128 *)&pxI;                                                        // x = *(float*)&i;
+    p = _mm_mul_ps(p, _mm_fmadd_ps(p, _mm_mul_ps(p, pHalfNeg), _ps_1p5));       // x = x * (1.5f - xHalf * x * x);
+
```
rrawther, line 177: We don't need empty line here

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -1439,6 +1669,130 @@ static inline __m128 atan2_ps( __m128 y, __m128 x )
...
+    // cut off denormalized stuff
+    x = _mm_max_ps(x, *(__m128 *)&_ps_min_norm_pos);
+
+    // part 1: x = frexpf(x, &e);
+    emm0 = _mm_srli_epi32(_mm_castps_si128(x), 23);
+
+    // keep only the fractional part
+    x = _mm_and_ps(x, *(__m128 *)&_ps_inv_mant_mask);
+    x = _mm_or_ps(x, *(__m128 *)&_ps_0p5);
+
```
rrawther, line 1751: I don't think we need to add empty lines in many places

src/include/func_specific/rng_seed_stream.hpp
```diff
@@ -5,7 +5,168 @@
...
+764129, 618740, 582752, 692507, 789955, 785851, 112601, 380578, 896709, 549581, 965817, 162355, 338288, 502799, 447091, 144368, 163130, 717741, 150842, 376320, 241741, 177798, 584469, 620725, 812482,
```
rrawther, line 8: Looks like the new values are not matching previous. Is this intentional

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1424,9 +1470,109 @@ __device__ __forceinline__ void rpp_hip_math_subtract24_const(d_float24 *src_f24
...
+}
+
+__device__ __forceinline__ float4 rpp_hip_math_inverse_sqrt4(float4 x_f4)
+{
+    float4 xHalf_f4 = (float4)0.5f * x_f4;
+    int4 i_i4 = *(int4 *)&x_f4;                                     // float bits in int
+    i_i4 = (int4) NEWTON_METHOD_INITIAL_GUESS - (i_i4 >> (int4)1);  // initial guess for Newton's method
+    x_f4 = *(float4 *)&i_i4;                                        // new bits to float
+    x_f4 = x_f4 * ((float4)1.5f - xHalf_f4 * x_f4 * x_f4);          // One round of Newton's method
+
```
rrawther, line 1531: remove the empty line

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1424,9 +1470,109 @@ __device__ __forceinline__ void rpp_hip_math_subtract24_const(d_float24 *src_f24
...
+    return x_f4;
+}
+
+__device__ __forceinline__ void rpp_hip_math_sqrt8(d_float8 *pix_f8, d_float8 *pixSqrt_f8)
+{
+    pixSqrt_f8->f4[0] = rpp_hip_math_inverse_sqrt4(pix_f8->f4[0]);
+    pixSqrt_f8->f4[1] = rpp_hip_math_inverse_sqrt4(pix_f8->f4[1]);
+
```
rrawther, line 1539: remove the empty line

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1424,9 +1470,109 @@ __device__ __forceinline__ void rpp_hip_math_subtract24_const(d_float24 *src_f24
...
+    pixSqrt_f24->f4[4] = one_f4 / pixSqrt_f24->f4[4];
+    pixSqrt_f24->f4[5] = one_f4 / pixSqrt_f24->f4[5];
+}
+
+__device__ __forceinline__ float rpp_hip_math_exp_lim256approx(float x)
+{
+  x = 1.0 + x * ONE_OVER_256;
+  x *= x; x *= x; x *= x; x *= x;
+  x *= x; x *= x; x *= x; x *= x;
+
```
rrawther, line 1568: remove empty line

rrawther (commented):
@r-abishek please address review comments

kiritigowda:
@paveltc - let me know if this passes all the unit tests.

paveltc:
@kiritigowda I commented yesterday that it does.

kiritigowda:
@r-abishek let me know when all review comments are resolved.

src/include/func_specific/rng_seed_stream.hpp
r-abishek: This one is intentional, the random seeds should be high enough to increase uniformity in randomization, but should be lower than the max value of the type. Reduced the size to avoid an out of bound issue.

r-abishek:
@rrawther Addressed all comments. Except the blank-line policy ones as per our chat.

rrawther (approved):
make sure tests are passing

r-abishek:
@kiritigowda Should be good to merge this one now.

PR #125, opened 2022-08-30
Resize interpolations HIP

r-abishek:
- Add vectorized HIP implementation for bilinear, bicubic, lanczos, triangular and gaussian interpolations on resize.
- Add corresponding test cases in test suite.


include/rppt_tensor_geometric_augmentations.h
```diff
@@ -129,6 +129,9 @@ RppStatus rppt_flip_gpu(RppPtr_t srcPtr, RpptDescPtr srcDescPtr, RppPtr_t dstPtr
 // *retval RPP_ERROR : Error
 
 RppStatus rppt_resize_host(RppPtr_t srcPtr, RpptDescPtr srcDescPtr, RppPtr_t dstPtr, RpptDescPtr dstDescPtr, RpptImagePatchPtr dstImgSizes, RpptInterpolationType interpolationType, RpptROIPtr roiTensorPtrSrc, RpptRoiType roiType, rppHandle_t rppHandle);
+#ifdef GPU_SUPPORT
+RppStatus rppt_resize_gpu(RppPtr_t srcPtr, RpptDescPtr srcDescPtr, RppPtr_t dstPtr, RpptDescPtr dstDescPtr, RpptImagePatchPtr dstImgSizes, RpptInterpolationType interpolationType, RpptROIPtr roiTensorPtrSrc, RpptRoiType roiType, rppHandle_t rppHandle);
```
rrawther, line 133: the function declarations are exactly same for host and gpu. Why do we need 2 different declarations?

src/include/hip/rpp_hip_common.hpp
```diff
@@ -88,6 +88,9 @@ struct RPPTensorFunctionMetaData
 #define ONE_OVER_255                    0.00392157f
 #define ONE_OVER_256                    0.00390625f
 #define SIX_OVER_360                    0.01666667f
+#define ONE_OVER_6                      0.1666666f
```
rrawther, line 91: I think too many of these definitions can cause rounding errors. Please use the following instead
const float ONE_OVER_SIX = 1.0f/6;
That way compiler will only do the calculation once and keep it as constant.

src/modules/hip/kernel/resize.hpp
```diff
@@ -0,0 +1,735 @@
...
+    locSrc_f16->f8[0].f4[1] = (locDst_f8x.f4[1] * (float4)wRatio) + wOffset_f4 + (float4)srcRoiPtr_i4->x;  // Compute src x locations in float for dst x locations [4-7]
+    locSrc_f16->f8[1].f4[0] = (locDst_f8y.f4[0] * (float4)hRatio) + hOffset_f4 + (float4)srcRoiPtr_i4->y;  // Compute src y locations in float for dst y locations [0-3]
+    locSrc_f16->f8[1].f4[1] = (locDst_f8y.f4[1] * (float4)hRatio) + hOffset_f4 + (float4)srcRoiPtr_i4->y;  // Compute src y locations in float for dst y locations [4-7]
+}
+
+__device__ void resize_roi_generic_srcloc_and_weight_hip_compute(int roiLoc, int dstLocation, float scale, int limit, int *srcLoc, float *weight, float offset, int srcStride)
+{
+    float srcLocationFloat = ((float) dstLocation) * scale + offset + (float)roiLoc;
+    int srcLocation = (int)ceilf(srcLocationFloat);
+    *weight = srcLocation - srcLocationFloat;
```
rrawther, line 31: You can combine 30 and 31 into 1 statement (*weight = ceilf(srcLocationFloat) - srcLocationFloat;)

src/modules/hip/kernel/resize.hpp
```diff
@@ -0,0 +1,735 @@
...
+
+    locSrc_f16->f8[0].f4[0] = (locDst_f8x.f4[0] * (float4)wRatio) + wOffset_f4 + (float4)srcRoiPtr_i4->x;  // Compute src x locations in float for dst x locations [0-3]
+    locSrc_f16->f8[0].f4[1] = (locDst_f8x.f4[1] * (float4)wRatio) + wOffset_f4 + (float4)srcRoiPtr_i4->x;  // Compute src x locations in float for dst x locations [4-7]
+    locSrc_f16->f8[1].f4[0] = (locDst_f8y.f4[0] * (float4)hRatio) + hOffset_f4 + (float4)srcRoiPtr_i4->y;  // Compute src y locations in float for dst y locations [0-3]
+    locSrc_f16->f8[1].f4[1] = (locDst_f8y.f4[1] * (float4)hRatio) + hOffset_f4 + (float4)srcRoiPtr_i4->y;  // Compute src y locations in float for dst y locations [4-7]
+}
+
+__device__ void resize_roi_generic_srcloc_and_weight_hip_compute(int roiLoc, int dstLocation, float scale, int limit, int *srcLoc, float *weight, float offset, int srcStride)
+{
+    float srcLocationFloat = ((float) dstLocation) * scale + offset + (float)roiLoc;
```
rrawther, line 29: Better to not name variables with their type in the name (xxxFloat)

src/modules/hip/kernel/resize.hpp
```diff
@@ -0,0 +1,735 @@
...
+
+    int4 srcRoi_i4 = *(int4 *)&roiTensorPtrSrc[id_z];
+    d_float16 locSrc_f16;
+    resize_roi_and_srclocs_hip_compute(&srcRoi_i4, &dstDimsWH, id_x, id_y, &locSrc_f16);
+
+    d_float8 dst_f8;
+    rpp_hip_interpolate8_bilinear_pln1(srcPtr + srcIdx, srcStridesNCH.z, &locSrc_f16, &srcRoi_i4, &dst_f8, false);
+    rpp_hip_pack_float8_and_store8(dstPtr + dstIdx, &dst_f8);
+
+    if (channelsDst == 3)
```
rrawther, line 103: is there a case for other values of channelsDst?

src/modules/hip/kernel/resize.hpp
```diff
@@ -0,0 +1,735 @@
...
+    resize_roi_generic_srcloc_and_weight_hip_compute(srcRoi_i4.y, id_y, hRatio, heightLimit, &srcLocationRowFloor, &rowWeight, hOffset, 1);
+
+    T *srcPtrTemp = srcPtr + (id_z * srcStridesNH.x);
+    float3 outPixel_f3 = (float3)0.0f;
+    float rowCoeffSum = 0.0f, colCoeffSum = 0.0f;
+    for(int j = 0; j < hKernelSize; j++)
+    {
+        int rowIndex = fminf(fmaxf((int)(srcLocationRowFloor + j), 0), heightLimit);
+        T *srcRowPtrsForInterp = srcPtrTemp + rowIndex * srcStridesNH.y;
+        rpp_hip_compute_interpolation_coefficient(interpolationType, (rowWeight - hRadius + j) * hScale , &rowCoeff);
```
rrawther, line 237: probably you can update the rowCoeffSum in the function itself instead of storing it in a separate variable

src/modules/hip/kernel/resize.hpp
```diff
@@ -0,0 +1,735 @@
...
+        for(int k = 0; k < wKernelSize; k++)
+        {
+            int colIndex = fminf(fmaxf((int)(srcLocationColumnFloor + k), 0), widthLimit);
+            rpp_hip_compute_interpolation_coefficient(interpolationType, (colWeight - wRadius + k) * wScale , &colCoeff);
+            colCoeffSum += colCoeff;
+            float3 coeff_f3 = (float3)(colCoeff * rowCoeff);
+            outPixel_f3 += (make_float3(srcRowPtrsForInterp[0][colIndex], srcRowPtrsForInterp[1][colIndex], srcRowPtrsForInterp[2][colIndex]) * coeff_f3);
+        }
+    }
+    outPixel_f3 *= (float3)(1 / (rowCoeffSum * colCoeffSum));
```
rrawther, line 328: check if there is a possibility of division by zero here

rrawther (commented):
@please address review comments
@asalmanp to review HIP kernels

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1689,6 +1728,104 @@ __device__ __forceinline__ float rpp_hip_rng_1_inverse_transform_sampling_f32(fl
...
+
+__device__ __forceinline__ void rpp_hip_compute_triangular_coefficient(float weight, float *coeff)
+{
+    *coeff = 1 - fabsf(weight);
+    *coeff = *coeff < 0 ? 0 : *coeff;
+}
+
+__device__ __forceinline__ void rpp_hip_compute_interpolation_coefficient(RpptInterpolationType interpolationType, float weight, float *coeff)
+{
+    switch (interpolationType)
```
asalmanp, line 1755: @r-abishek as discussed, please see if a combination of the templated kernels and function pointers can be used to avoid using the switch cases. Thanks.

asalmanp (changes requested):

src/modules/hip/kernel/resize.hpp
sampath1117: We are using the rounded value (srcLocation) from Line 30 in [Line 32](https://github.com/r-abishek/rpp/blob/ar/opt_resize_hip/src/modules/hip/kernel/resize.hpp#L32) 
*srcLoc = ((srcLocation > limit) ? limit : srcLocation) * srcStride;

if we combine Line 30 and Line 31
we need to call ceilf 2 additional times in Line 32 as below which can be costly
*srcLoc = (((int)ceilf(srcLocation) > limit) ? limit : ceilf(srcLocation)) * srcStride;

src/modules/hip/kernel/resize.hpp
sampath1117: Currently we are only handling cases for channelsDst =1 (PLN1) and channelsDst = 3 (PLN3)

src/modules/hip/kernel/resize.hpp
sampath1117: @rrawther 
There is a possibility of 0 as some interpolation functions will return 0 as coeff based on the weights being passed as input to interpolation functions
https://github.com/r-abishek/rpp/blob/ar/opt_resize_hip/src/include/hip/rpp_hip_common.hpp#L1750
https://github.com/r-abishek/rpp/blob/ar/opt_resize_hip/src/include/hip/rpp_hip_common.hpp#L1739

if we initialize the value to  very small number !=0  then 1/(very small number) will become a large value which when multiplied by outPixel_f3 makes it as a large number / can make it out of range

Is it fine if we can add two branch statements like below to handle this case?
`rowCoeffSum = (rowCoeffSum == 0.0f) ? 1.0f : rowCoeffSum;
    colCoeffSum = (colCoeffSum == 0.0f) ? 1.0f : colCoeffSum;
    outPixel_f3 *= (float3)(1 / (rowCoeffSum * colCoeffSum));`

src/include/hip/rpp_hip_common.hpp

src/modules/hip/kernel/resize.hpp
r-abishek: Done

include/rppt_tensor_geometric_augmentations.h
r-abishek: We could have it in a single api, but this would ideally be a common change in a separate PR for all functions. We would need changes in MIVisionX too so it doesn't break

src/include/hip/rpp_hip_common.hpp
r-abishek: Verified this one. It passes training, and templated switch case performs too. Merged with this PR branch now.

paveltc:
@r-abishek @rrawther @asalmanp @kiritigowda This PR passes all of the relevant unit tests.

PR #126, opened 2022-09-21
Crop Mirror  Normalize updates

r-abishek:
- CMN variants U8 -> F32 and  U8 -> F16 added
- Support for normalization per R, G, B channel added


src/modules/rppt_tensor_geometric_augmentations.cpp
```diff
@@ -153,6 +153,33 @@ RppStatus rppt_crop_mirror_normalize_host(RppPtr_t srcPtr,
                                                 roiType,
                                                 layoutParams);
     }
+    else if ((srcDescPtr->dataType == RpptDataType::U8) && (dstDescPtr->dataType == RpptDataType::F32))
+    {
+        crop_mirror_normalize_u8_f32_host_tensor(static_cast<Rpp8u*>(srcPtr) + srcDescPtr->offsetInBytes,
+                                                 srcDescPtr,
+                                                 (Rpp32f*) (static_cast<Rpp8u*>(dstPtr) + dstDescPtr->offsetInBytes),
+                                                 dstDescPtr,
+                                                 meanTensor,
```
rrawther, line 162: are the mean and std values different for each image in the tensor?

src/modules/cpu/kernel/crop_mirror_normalize.hpp
```diff
@@ -81,9 +82,9 @@ RppStatus crop_mirror_normalize_u8_u8_host_tensor(Rpp8u *srcPtr,
 
                     for (; vectorLoopCount < bufferLength; vectorLoopCount += 3)
                     {
```
rrawther, line 84: adjust spacing here

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -688,6 +733,142 @@ inline void rpp_normalize48(__m128 *p)
 
 // AVX loads and stores
 
+inline void rpp_load48_u8pkd3_to_f32pkd3_avx(Rpp8u *srcPtr, __m256 *p)
+{
+    __m128i px[4];
```
rrawther, line 738: It is hard to verify all these AVX intrinsics in the core review. I suggest having a mechanism in the code which verify the AVX code automatically. Basically we need to run reference c code, store output in some temp buffers, run AVX code and compare if it exactly matching the reference c implementation

src/modules/cpu/kernel/crop_mirror_normalize.hpp
```diff
@@ -131,9 +132,9 @@ RppStatus crop_mirror_normalize_u8_u8_host_tensor(Rpp8u *srcPtr,
...
```
rrawther, line 133: for all for loops move the bracket to previous line

src/modules/cpu/kernel/crop_mirror_normalize.hpp
```diff
@@ -131,9 +132,9 @@ RppStatus crop_mirror_normalize_u8_u8_host_tensor(Rpp8u *srcPtr,
                     for (; vectorLoopCount < bufferLength; vectorLoopCount += 3)
```
rrawther, line 132: how is the AVX code and C code has the same vector loop increment

src/modules/hip/kernel/crop_mirror_normalize.hpp
```diff
@@ -89,66 +106,78 @@ __global__ void crop_mirror_normalize_pln_tensor(T *srcPtr,
...
 
     uint srcIdx;
-    d_float8 pix_f8, cmnParams_f8;
     uint dstIdx = (id_z * dstStridesNCH.x) + (id_y * dstStridesNCH.z) + id_x;
-
-    cmnParams_f8.f4[0] = (float4)meanTensor[id_z];
-    cmnParams_f8.f4[1] = (float4)(1 / stdDevTensor[id_z]);
+    int incrementPerImage = id_z * channelsDst;
+    d_float8 pix_f8, cmnParams_f8;
+    cmnParams_f8.f4[0] = (float4)meanTensor[incrementPerImage];          // Get mean for R channel
```
rrawther, line 112: Looks like there is inconsistency between c and hip. Is the mean and std values different for each image in a tensor?

src/modules/hip/kernel/crop_mirror_normalize.hpp
```diff
@@ -89,66 +106,78 @@ __global__ void crop_mirror_normalize_pln_tensor(T *srcPtr,
...
+    cmnParams_f8.f4[1] = (float4)(1 / stdDevTensor[incrementPerImage]);  // Get (1 / stdDev) for R channel
```
rrawther, line 113: it is better to calculate this at higher level and pass 1/std value to the kernel

rrawther (commented):
let us do a code walkthrough of this PR. Many things are not clear for me

src/modules/rppt_tensor_geometric_augmentations.cpp
sampath1117: Hi Rajy,
Yes, the mean and std values can be different for each image in the tensor

src/modules/cpu/kernel/crop_mirror_normalize.hpp
sampath1117: The avx code has an increment by value (vectorIncrement = 48) and c code has an increment by 3
https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/126/files#diff-c293c511d43aee8d44e0e51e1aabf688ee9ec991096f27cd19f4dc4f99c8cb25R119

We are processing 16 RGB pixels (48 pixel values in total) per iteration in the AVX loop along width and remaining pixels along width are processed in c loop

src/modules/hip/kernel/crop_mirror_normalize.hpp
sampath1117: Yes mean and std values can be different for each image in tensor

src/modules/hip/kernel/crop_mirror_normalize.hpp
```diff
@@ -1,33 +1,45 @@
 #include <hip/hip_runtime.h>
 #include "rpp_hip_common.hpp"
 
-__device__ void cmn_hip_compute(uchar *srcPtr, d_float8 *pix_f8, d_float8 *cmnParams_f8)
+__device__ void cmn_hip_compute(uchar *srcPtr, float *dstPtr, d_float8 *pix_f8, d_float8 *cmnParams_f8)
+{
+    pix_f8->f4[0] = rpp_hip_pixel_check_0to1((pix_f8->f4[0] - cmnParams_f8->f4[0]) * cmnParams_f8->f4[1] * (float4) ONE_OVER_255);
```
rrawther, line 6: @asalmanp : Can you verify if we really need this pixel check. Can we safely assume the output will be in 0-1 range?

rrawther (commented):
@asalmanp : Can you verify the HIP kernel for this?
@paveltc to run unit tests
@sampath1117 : Please address the issues we have discussed in code_review

src/modules/hip/kernel/crop_mirror_normalize.hpp
asalmanp: @rrawther we need to know the range of the pix_f8 and cmnParams_f8 to see if we can make the assumption that the output is between 0 to 1. @r-abishek can you verify the range of these variables?

r-abishek: @rrawther @asalmanp All rpp tensor functions till now ensure all outputs for f16/f32 are in 0-1, for u8 are in 0-255 and i8 are in -128-127. Just for CMN, we see that this cannot be done (can't safely assume that the output will be between 0-1 since it does a (src - mean) / stdDev and the values could go out of range.

However, please note that in a few more changes being done to this PR, we will be removing the rpp_hip_pixel_check_0to1() for CMN since the math from the PyTorch training was verified and the outputs seem to do no range check. (So, output values are going outside the 0-1 range in PyTorch version). 

To match that, RPP will have to have an exception for just the CMN, that we don't check and saturate boundaries for all CMN functions. Please let us know any thoughts

rrawther (commented):
@r-abishek @sampath1117 : Please address the review comments for formatting

src/modules/hip/kernel/crop_mirror_normalize.hpp
r-abishek: 1/stdDev has been moved out now

src/modules/cpu/kernel/crop_mirror_normalize.hpp
r-abishek: Extra line removed

r-abishek:
@rrawther This PR is verified to be compatible on training with both HIP/HOST.

paveltc:
@rrawther This passes all unit tests.

src/modules/cpu/kernel/crop_mirror_normalize.hpp
r-abishek: addressed in code walkthrough call. Common change in a separate PR after these PR merges.

src/modules/hip/kernel/crop_mirror_normalize.hpp
rrawther: Ok. We need to analyze each function and see if the pixel range check is needed.

rrawther (approved):
@paveltc: make sure the tests are passing
@r-abishek : please remove the WIP for PRs which are ready to be merged and complete

PR #130, opened 2022-11-29
Non linear blend Tensor impl - HOST

r-abishek:
- Adds Non linear blend tensor implementation on HOST

### DIFF
--- a/src/modules/cpu/kernel/non_linear_blend.hpp
+++ b/src/modules/cpu/kernel/non_linear_blend.hpp
@@ -1,0 +1,1144 @@
+#include "rppdefs.h"
+#include "rpp_cpu_simd.hpp"
+#include "rpp_cpu_common.hpp"
+
+// -------------------- non_linear_blend host helpers --------------------
+
+inline void compute_non_linear_blend_48_host(__m256 *p1, __m256 *p2, __m256 &pMultiplier, __m256 &pILocComponent, __m256 &pJLocComponent)
+{
+    __m256 pGaussianValue;
+    pGaussianValue = fast_exp_avx(_mm256_fmadd_ps(_mm256_mul_ps(pJLocComponent, pJLocComponent), pMultiplier, pILocComponent));
+    p1[0] = _mm256_fmadd_ps(_mm256_sub_ps(p1[0], p2[0]), pGaussianValue, p2[0]);    // non_linear_blend adjustment
+    p1[2] = _mm256_fmadd_ps(_mm256_sub_ps(p1[2], p2[2]), pGaussianValue, p2[2]);    // non_linear_blend adjustment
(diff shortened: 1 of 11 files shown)


paveltc:
@rrawther @kiritigowda @r-abishek This PR passes all relevant unit tests.

src/modules/cpu/kernel/non_linear_blend.hpp
```diff
@@ -0,0 +1,1577 @@
...
+    omp_set_dynamic(0);
+#pragma omp parallel for num_threads(dstDescPtr->n)
+    for(int batchCount = 0; batchCount < dstDescPtr->n; batchCount++)
+    {
+        RpptROI roi;
+        RpptROIPtr roiPtrInput = &roiTensorPtrSrc[batchCount];
+        compute_roi_validation_host(roiPtrInput, &roi, &roiDefault, roiType);
+
+        Rpp32f stdDev = stdDevTensor[batchCount];
+        Rpp32f multiplier = -0.5f / (stdDev * stdDev);
```
rrawther, line 74: probably need to check std value to avoid division by 0

utilities/rpp-performancetests/HOST_NEW/Tensor_host_pkd3.cpp
```diff
@@ -1156,6 +1159,57 @@ int main(int argc, char **argv)
 
             break;
         }
+        case 30:
```
rrawther, line 1162: looks like case 30 is repeated here

utilities/rpp-performancetests/HOST_NEW/Tensor_host_pln1.cpp
```diff
@@ -1147,6 +1150,57 @@ int main(int argc, char **argv)
 
             break;
         }
+        case 30:
```
rrawther, line 1153: case 30 repeated in all test cases

rrawther (commented):

utilities/rpp-performancetests/HOST_NEW/Tensor_host_pkd3.cpp
r-abishek: This one is actually needed, we need the name of the functionality being run first - To decide all the u8/f32/f16/i8 or if the case being run is special (like has interpolationType or filterKernelSize). After all params are locked to the variant being run (with common code), there is another switch case to actually run the RPP call.

utilities/rpp-performancetests/HOST_NEW/Tensor_host_pln1.cpp
r-abishek: Currently same in all test suite

kiritigowda:
@r-abishek can you resolve conversations on the PR ?

rrawther (approved):

PR #131, opened 2022-11-29
Non linear blend Tensor impl - HIP

r-abishek:
- Adds Non linear blend tensor implementation on HIP

### DIFF
--- a/src/modules/hip/kernel/non_linear_blend.hpp
+++ b/src/modules/hip/kernel/non_linear_blend.hpp
@@ -1,0 +1,286 @@
+#include <hip/hip_runtime.h>
+#include "rpp_hip_common.hpp"
+
+__device__ void non_linear_blend_gaussian_hip_compute(float &multiplier, int2 &halfDimsWH_i2, int2 &idXY_i2, d_float8 *gaussianValue_f8)
+{
+    float rowLocComponent;
+    rowLocComponent = idXY_i2.y - halfDimsWH_i2.y;
+    rowLocComponent = rowLocComponent * rowLocComponent * multiplier;
+
+    float4 rowLocComponent_f4 = (float4)rowLocComponent;
+    float4 multiplier_f4 = (float4)multiplier;
+
+    d_float8 colLocComponent_f8;
+    colLocComponent_f8.f4[0] = make_float4(idXY_i2.x, idXY_i2.x + 1, idXY_i2.x + 2, idXY_i2.x + 3);
+    colLocComponent_f8.f4[1] = colLocComponent_f8.f4[0] + (float4)4;
+    colLocComponent_f8.f4[0] -= (float4)halfDimsWH_i2.x;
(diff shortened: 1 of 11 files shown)


paveltc:
@rrawther @kiritigowda @r-abishek This PR passes all relevant unit tests.

src/modules/hip/kernel/non_linear_blend.hpp
```diff
@@ -0,0 +1,286 @@
+#include <hip/hip_runtime.h>
+#include "rpp_hip_common.hpp"
+
+__device__ void non_linear_blend_gaussian_hip_compute(float &multiplier, int2 &halfDimsWH_i2, int2 &idXY_i2, d_float8 *gaussianValue_f8)
+{
+    float rowLocComponent;
+    rowLocComponent = idXY_i2.y - halfDimsWH_i2.y;
+    rowLocComponent = rowLocComponent * rowLocComponent * multiplier;
```
rrawther, line 8: you can use *= here

src/modules/hip/kernel/non_linear_blend.hpp
```diff
@@ -0,0 +1,286 @@
...
+
+    if ((id_y >= roiTensorPtrSrc[id_z].xywhROI.roiHeight) || (id_x >= roiTensorPtrSrc[id_z].xywhROI.roiWidth))
+    {
+        return;
+    }
+
+    uint srcIdx = (id_z * srcStridesNH.x) + ((id_y + roiTensorPtrSrc[id_z].xywhROI.xy.y) * srcStridesNH.y) + ((id_x + roiTensorPtrSrc[id_z].xywhROI.xy.x) * 3);
+    uint dstIdx = (id_z * dstStridesNH.x) + (id_y * dstStridesNH.y) + id_x * 3;
+
+    float multiplier = -0.5f / (stdDev[id_z] * stdDev[id_z]);
```
rrawther, line 59: Is there a way to avoid this floating point division inside the kernel?

rrawther (commented):
address review comments

src/modules/hip/kernel/non_linear_blend.hpp
r-abishek: Done

src/modules/hip/kernel/non_linear_blend.hpp
r-abishek: It's hard since stdDev is supplied by the user. In the short term, we could always do a host loop outside computing (-0.5f/(stdDev*stdDev)), followed by conversion to device mem.
But in the long term, if RPP will be expecting uniformly all params to be in GPU mem too, then this has to ideally be done inside the kernel.

rrawther: Yes. I think we need to figure a way to do this more efficiently in future version

rrawther (approved):

kiritigowda:
@r-abishek can you resolve the conflict?

r-abishek:
@kiritigowda Resolved

PR #132, opened 2022-11-30
Rotate Tensor HOST and HIP

r-abishek:
- Adds rotate tensor API for HOST and HIP
- Adds relevant unit and perf tests


src/modules/rppt_tensor_geometric_augmentations.cpp
```diff
@@ -1269,4 +1394,84 @@ return RPP_SUCCESS;
...
+                          RpptROIPtr roiTensorPtrSrc,
+                          RpptRoiType roiType,
+                          rppHandle_t rppHandle)
+{
+#ifdef HIP_COMPILE
+    if ((interpolationType != RpptInterpolationType::BILINEAR) && (interpolationType != RpptInterpolationType::NEAREST_NEIGHBOR))
+        return RPP_ERROR_NOT_IMPLEMENTED;
+    
+    // Compute affine transformation matrix from rotate angle
+    Rpp32f *affineTensor = rpp::deref(rppHandle).GetInitHandle()->mem.mcpu.tempFloatmem;
```
rrawther, line 1414: why is this stored in cpu mem for HIP backend?

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pln1.cpp
```diff
@@ -1267,6 +1271,66 @@ int main(int argc, char **argv)
...
+                roiTensorPtrSrc[i].ltrbROI.lt.y = 30;
+                roiTensorPtrSrc[i].ltrbROI.rb.x = 210;
+                roiTensorPtrSrc[i].ltrbROI.rb.y = 210;
+                dstImgSizes[i].width = roiTensorPtrSrc[i].ltrbROI.rb.x - roiTensorPtrSrc[i].ltrbROI.lt.x + 1;
+                dstImgSizes[i].height = roiTensorPtrSrc[i].ltrbROI.rb.y - roiTensorPtrSrc[i].ltrbROI.lt.y + 1;
+            }
+            roiTypeSrc = RpptRoiType::LTRB;
+            roiTypeDst = RpptRoiType::LTRB;*/
+
+            hipMemcpy(d_roiTensorPtrSrc, roiTensorPtrSrc, images * sizeof(RpptROI), hipMemcpyHostToDevice);
```
rrawther, line 1311: you can use a pinned memory instead using hipHostMalloc()

rrawther (commented):

src/modules/rppt_tensor_geometric_augmentations.cpp
fiona-gladwin: The hip_exec_warp_affine_tensor internally does a hipMemcopy [here ](https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/blob/a364d41a299b4d017c080dce7ed5233aed7ca9a4/src/modules/hip/kernel/warp_affine.hpp#L338)for the affine tensor buffer.
Hence we are computing the affine tensor and storing it in a pre existing memory in the handle and reusing the existing warp affine kernels for Rotate augmentation.

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pln1.cpp
fiona-gladwin: Modified HIP test suite to use pinned memory for ROI and dstImageSizes

kiritigowda:
@LakshmiKumar23 let me know when this PR is ready?

utilities/rpp-performancetests/HIP_NEW/Tensor_hip_pln1.cpp
```diff
@@ -94,9 +94,9 @@ int main(int argc, char **argv)
     unsigned int outputFormatToggle = atoi(argv[4]);
     int test_case = atoi(argv[5]);
 
-    bool additionalParamCase = (test_case == 8 || test_case == 21 || test_case == 24 || test_case == 40 || test_case == 41 || test_case == 49);
+    bool additionalParamCase = (test_case == 8 || test_case == 21 || test_case == 23 || test_case == 24 || test_case == 40 || test_case == 41 || test_case == 49);
```
rrawther, line 97: We have so many file changes for tests which is changed for every new kernel support of RPP. It is very hard to review. We need to take a pause and make combine all these test scrips into single file which can be controlled with arguments. Plus it will increase the productivity of test creation as well since you don't need to repeat this changes to all

rrawther (commented):
@paveltc  to run tests

kiritigowda:
@r-abishek is this ready to be merged?

LakshmiKumar23:
@kiritigowda This is ready to merge

PR #138, opened 2023-02-09
Fix HIP Tensor Resize Kernel

snehaa8:
Fixes Triangular Interpolation


rrawther (commented):
@snehaa8: not sure who this is. But it is already a closed PR. Making commits into it will not get merged into mainline

PR #144, opened 2023-04-20
Sample for HIP training course

LakshmiKumar23:
Using only brightness as an example from the test suite for hip training course.


asalmanp (commented):
@LakshmiKumar23 can you please move the entire examples folder under utilities folder?

LakshmiKumar23:
> @LakshmiKumar23 can you please move the entire examples folder under utilities folder?

@asalmanp Also added a REAME and fixed codacy issues in the same commit

PR #152, opened 2023-05-23
LUT tensor on HOST and HIP

r-abishek:
Adds LUT tensor support for U8->U8, U8->F16, U8->F32 and I8->I8 variants
Adds LUT tensor support for PKD3, PLN3, PLN1 with toggle
Adds test case for LUT in new test_suite for QA testing, unit, performance tests
Adds golden outputs for LUT with match on host and hip


include/rppt_tensor_color_augmentations.h
```diff
@@ -199,6 +199,26 @@ RppStatus rppt_contrast_host(RppPtr_t srcPtr, RpptDescPtr srcDescPtr, RppPtr_t d
...
+// *param[in] dstDescPtr destination tensor descriptor
+// *param[in] lutPtr lut Array containing a single integer look up table of length 65536, to be used for all images in the batch
+// *param[in] roiTensorSrc ROI data for each image in source tensor (2D tensor of size batchSize * 4, in either format - XYWH(xy.x, xy.y, roiWidth, roiHeight) or LTRB(lt.x, lt.y, rb.x, rb.y))
+// *param[in] roiType ROI type used (RpptRoiType::XYWH or RpptRoiType::LTRB)
+// *returns a  RppStatus enumeration.
+// *retval RPP_SUCCESS : succesful completion
+// *retval RPP_ERROR : Error
+
+RppStatus rppt_lut_host(RppPtr_t srcPtr, RpptDescPtr srcDescPtr, RppPtr_t dstPtr, RpptDescPtr dstDescPtr, RppPtr_t lutPtr, RpptROIPtr roiTensorPtrSrc, RpptRoiType roiType, rppHandle_t rppHandle);
+#ifdef GPU_SUPPORT
```
rrawther, line 218: The host and gpu functions look exactly same except with name change. We need to combine them into one api call and indicate some other way to differentiate so we can get rid of GPU_SUPPORT flag

src/modules/cpu/kernel/lut.hpp
```diff
@@ -0,0 +1,569 @@
...
+            {
+                Rpp8u *srcPtrTemp, *dstPtrTempR, *dstPtrTempG, *dstPtrTempB;
+                srcPtrTemp = srcPtrRow;
+                dstPtrTempR = dstPtrRowR;
+                dstPtrTempG = dstPtrRowG;
+                dstPtrTempB = dstPtrRowB;
+
+                int vectorLoopCount = 0;
+                for (; vectorLoopCount < bufferLength; vectorLoopCount += 3)
+                {
```
rrawther, line 52: all of these are pure c code. Are we supposed to add SSE/AVX code?

src/modules/hip/kernel/lut.hpp
```diff
@@ -0,0 +1,250 @@
...
+    {
+        return;
+    }
+
+    uint srcIdx = (id_z * srcStridesNH.x) + ((id_y + roiTensorPtrSrc[id_z].xywhROI.xy.y) * srcStridesNH.y) + (id_x + roiTensorPtrSrc[id_z].xywhROI.xy.x * 3);
+    uint dstIdx = (id_z * dstStridesNH.x) + (id_y * dstStridesNH.y) + id_x;
+
+    d_float8 src_f8, dst_f8;
+
+    rpp_hip_load8_and_unpack_to_float8(srcPtr + srcIdx, &src_f8);
```
rrawther, line 59: the kernel is using computations assuming the data_type of Src and Dst whereas the types are templated.
How are we planning to support other data_types for  Src and Dst. Same true for other kernel implementations here

src/modules/rppt_tensor_color_augmentations.cpp
```diff
@@ -605,6 +605,67 @@ RppStatus rppt_contrast_host(RppPtr_t srcPtr,
...
+    {
+        lut_i8_i8_host_tensor(static_cast<Rpp8s*>(srcPtr) + srcDescPtr->offsetInBytes,
+                              srcDescPtr,
+                              static_cast<Rpp8s*>(dstPtr) + dstDescPtr->offsetInBytes,
+                              dstDescPtr,
+                              static_cast<Rpp8s*>(lutPtr),
+                              roiTensorPtrSrc,
+                              roiType,
+                              layoutParams);
+    }
```
rrawther, line 664: what about other datatypes? Please return failure with error message for unsupported datatypes

utilities/rpp-unittests/HIP_NEW/Tensor_hip_pkd3.cpp
```diff
@@ -2364,7 +2364,7 @@ int main(int argc, char **argv)
         count += dstDescPtr->strides.nStride;
```
rrawther, line 2365: coding of many stuff in this file is not done very good. Are we going to keep this file

utilities/rpp-unittests/HIP_NEW/Tensor_hip_pkd3.cpp
```diff
@@ -2364,7 +2364,7 @@ int main(int argc, char **argv)
...
 
         char temp[1000];
-        strcpy(temp, dst);
+        strcpy(temp, "dst");
```
rrawther, line 2367: need to use std::strings instead of old c style coding

utilities/test_suite/HIP/Tensor_hip.cpp
```diff
@@ -521,6 +521,46 @@ int main(int argc, char **argv)
...
+            testCaseName = "lut";
+
+            Rpp32f *lutBuffer;
+            hipHostMalloc(&lutBuffer, 65536 * sizeof(Rpp32f));
+            hipMemset(lutBuffer, 0, 65536 * sizeof(Rpp32f));
+            Rpp8u *lut8u = reinterpret_cast<Rpp8u *>(lutBuffer);
+            Rpp16f *lut16f = reinterpret_cast<Rpp16f *>(lutBuffer);
+            Rpp32f *lut32f = reinterpret_cast<Rpp32f *>(lutBuffer);
+            Rpp8s *lut8s = reinterpret_cast<Rpp8s *>(lutBuffer);
+            if (inputBitDepth == 0)
```
rrawther, line 535: Very inefficient ways of if else .. switch statements, need to revisit

utilities/test_suite/HOST/runTests.py
```diff
@@ -144,7 +144,7 @@ def rpp_test_suite_parser_and_validator():
 subprocess.call(["./testAllScript.sh", srcPath1, args.input_path2, str(testType), str(numIterations), str(qaMode), str(decoderType), str(preserveOutput), " ".join(caseList)])  # nosec
 
 # print the results of qa tests
-supportedCaseList = ['0', '2', '4', '13', '31', '36', '38']
+supportedCaseList = ['0', '2', '4', '13', '31', '34', '36', '38']
```
rrawther, line 147: constant tables has to be defined outside the function

rrawther (changes requested):

src/modules/hip/kernel/lut.hpp
r-abishek: The rpp_hip_load8_and_unpack_to_float8() function actually takes any datatype and converts to float.
Also, as we discussed lut is only useful in U8/I8 inputs, but the outputs can be U8/I8/F16/F32 for which we have support.
A [256] length or [65536] length look up table could be provided by the user.

src/modules/rppt_tensor_color_augmentations.cpp
r-abishek: src as F16/F32 is not supported since it won't make sense, and we'll add the error message.

utilities/rpp-unittests/HIP_NEW/Tensor_hip_pkd3.cpp
r-abishek: This file is to be deleted once the new test_suite is ready, @snehaa8 We need to remove any updates to these files.

utilities/test_suite/HIP/Tensor_hip.cpp
r-abishek: We went over this internally, and these buffers are just sample for the user running the unit test. But I felt it was hard to separately initialize 4 buffers. This is what we need:
One (uchar) buffer of length [256] - with values (256 - i)
One (float) buffer of length [256] - with values (256 - i)/255
One (half) buffer of length [256] - with values (256 - i)/255
One (signed char) buffer of length [256] - with values (128 - i)
We can discuss suggestions from your side Rajy

include/rppt_tensor_color_augmentations.h
snehaa8: Sure Rajy, we will explore on this.

utilities/rpp-unittests/HIP_NEW/Tensor_hip_pkd3.cpp
snehaa8: Yeah Abishek, missed removing this during cleanup. I will clean this up.

utilities/rpp-unittests/HIP_NEW/Tensor_hip_pkd3.cpp
snehaa8: I will clean this up Rajy.

src/modules/cpu/kernel/lut.hpp
snehaa8: Yeah Rajy, these are purely C codes. We didn't feel SSE/AVX might give improvements as its direct lookup followed by replacement. Please let us know on your suggestions too.

utilities/test_suite/HOST/runTests.py
snehaa8: Sure Rajy, we will address this in Testsuite PR.

src/modules/rppt_tensor_color_augmentations.cpp
r-abishek: error code added

include/rppt_tensor_color_augmentations.h
r-abishek: We'll have to discuss and look at this in a separate PR. It will have a common change for every functionality.

kiritigowda:
@rrawther is this ready to be merged?

utilities/test_suite/HOST/Tensor_host.cpp
```diff
@@ -511,6 +511,44 @@ int main(int argc, char **argv)
...
+                Rpp8u *lut8u = reinterpret_cast<Rpp8u *>(lutBuffer);
+                Rpp16f *lut16f = reinterpret_cast<Rpp16f *>(lutBuffer);
+                Rpp32f *lut32f = reinterpret_cast<Rpp32f *>(lutBuffer);
+                Rpp8s *lut8s = reinterpret_cast<Rpp8s *>(lutBuffer);
+                if (inputBitDepth == 0)
+                    for (j = 0; j < 256; j++)
+                        lut8u[j] = (Rpp8u)(255 - j);
+                else if (inputBitDepth == 3)
+                    for (j = 0; j < 256; j++)
+                        lut16f[j] = (Rpp16f)((255 - j) * ONE_OVER_255);
```
rrawther, line 528: probably you should do division here so we don't loose precision

rrawther (approved):
@paveltc to run tests

kiritigowda:
@rrawther can you look at the final commits after the approval. @paveltc let me know when this PR is good to merge.

utilities/test_suite/HOST/Tensor_host.cpp
r-abishek: This is actually just a sample case in the test suite for users to see some example. The user could always modify to give any LUT of their choice, its not part of the functionality

paveltc:
@r-abishek Will the white image results be fixed in this PR or a future one? I remember you can't add anything to this PR because it's already approved.

r-abishek:
@paveltc The fix is now reflected in this PR#152 and we see the unit tests passing.  Its a minor fix in the test suite.
Github just took a while before the changes showed up here on the PR.

PR #171, opened 2023-06-29
openCL build fix

r-abishek:
Fixes build fail on openCL @kiritigowda @LakshmiKumar23

### DIFF
--- a/src/modules/cl/handleocl.cpp
+++ b/src/modules/cl/handleocl.cpp
@@ -29,6 +29,7 @@
 #include <unistd.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


kiritigowda (approved):

kiritigowda:
[100%] Linking CXX shared library lib/librpp.so
[100%] Built target rpp
+ sudo make install
Consolidate compiler generated dependencies of target addkernels
[  5%] Built target addkernels
Consolidate compiler generated dependencies of target modules
[ 98%] Built target modules
[100%] Built target rpp

PR #52, opened 2021-04-20
Enable static build option for statistical_operations

r-abishek:

PR #179, opened 2023-07-13
RPP Tensor Image Sum for HOST and HIP

r-abishek:
- Adds support for reduction image sum kernels on tensor HOST and vectorized tensor HIP
- Adds test case for image sum HOST and HIP kernels in new test suite
- Adds respective QA tests, unit tests and performance tests


include/rppt_tensor_statistical_operations.h
```diff
@@ -0,0 +1,53 @@
+/*
+Copyright (c) 2019 - 2022 Advanced Micro Devices, Inc. All rights reserved.
```
rrawther, line 2: change the data

include/rppt_tensor_statistical_operations.h
```diff
@@ -0,0 +1,53 @@
...
+// *param[in] srcDescPtr source tensor descriptor (srcDescPtr->w can be a maximum of 3840, srcDescPtr->h can be a maximum of 2160)
+// *param[out] imageSumArr destination array of minimum length (srcPtr->n * srcPtr->c)
+// *param[in] imageSumArrLength length of provided destination array (minimum length = srcPtr->n * srcPtr->c)
+// *param[in] roiTensorSrc ROI data for each image in source tensor (2D tensor of size batchSize * 4, in either format - XYWH(xy.x, xy.y, roiWidth, roiHeight) or LTRB(lt.x, lt.y, rb.x, rb.y))
+// *param[in] roiType ROI type used (RpptRoiType::XYWH or RpptRoiType::LTRB)
+// *returns a  RppStatus enumeration.
+// *retval RPP_SUCCESS : succesful completion
+// *retval RPP_ERROR : Error
+
+RppStatus rppt_image_sum_host(RppPtr_t srcPtr, RpptDescPtr srcDescPtr, RppPtr_t imageSumArr, Rpp32u imageSumArrLength, RpptROIPtr roiTensorPtrSrc, RpptRoiType roiType, rppHandle_t rppHandle);
```
rrawther, line 45: why are we calling tensor api with "image"

src/include/func_specific/reduction.hpp
```diff
@@ -0,0 +1,33 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#ifndef RPP_REDUCTION_H
+#define RPP_REDUCTION_H
+
+#include <rppdefs.h>
+
+#define REDUCTION_MAX_WIDTH 3840
```
rrawther, line 28: How do we choose these dimensions as max?

src/modules/cpu/kernel/image_sum.hpp
```diff
@@ -0,0 +1,846 @@
+#include "rppdefs.h"
+#include "rpp_cpu_simd.hpp"
+#include "rpp_cpu_common.hpp"
+#include "reduction.hpp"
+
+RppStatus image_sum_u8_u8_host_tensor(Rpp8u *srcPtr,
```
rrawther, line 6: this name is confusing for the api. please rename it to more meaningful name. Please don't combine image and tensor names in a single api unless it is converting from image to tensor or veceversa.
For exampe (tensor_sum_u32_u8_host). Also why are we outputting the sum into a float 32. It will loose precision

src/modules/hip/hip_tensor_statistical_operations.hpp
```diff
@@ -0,0 +1,25 @@
+/*
+Copyright (c) 2019 - 2022 Advanced Micro Devices, Inc. All rights reserved.
```
rrawther, line 2: update date, applies to all such comments in other files

src/modules/hip/kernel/image_sum.hpp
```diff
@@ -0,0 +1,481 @@
+#include <hip/hip_runtime.h>
+#include "rpp_hip_common.hpp"
+#include "reduction.hpp"
+
+// -------------------- Set 0 - Reduction Stage 2 --------------------
+
+__global__ void image_sum_grid_result_tensor(float *srcPtr,
```
rrawther, line 7: we need to revisit. Adding all u8 values and putting it into float output will loose precision. Also where is this kernel used. If we are outputing into float, the input has to be converted and normalized to float (0-1.f ) range before summing

rrawther (changes requested):
I am not sure if the implementation is correct for these kernels. So we need to revisit

rrawther:
@r-abishek  Can you please addess the review comments

include/rppt_tensor_statistical_operations.h
r-abishek: Done for all instances

include/rppt_tensor_statistical_operations.h
r-abishek: Done - changed to tensor_sum for clarity

src/include/func_specific/reduction.hpp
r-abishek: This comes from the max size we can accommodate for block sums in the second vectorized hip kernel. Might be better to discuss on call.

src/modules/cpu/kernel/image_sum.hpp
r-abishek: Done - all names are now tensor_sum

src/modules/hip/hip_tensor_statistical_operations.hpp
r-abishek: Done

src/modules/hip/kernel/image_sum.hpp
r-abishek: Changed - added Rpp32u (uint) and Rpp32s (int) sums support

kiritigowda:
@r-abishek any update on this?

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1429,6 +1466,81 @@ __device__ __forceinline__ void rpp_hip_layouttoggle24_pln3_to_pkd3(T *pixpln3Pt
...
+{
+    d_schar8_s src_sc8;
+    *(d_schar8_s *)&src_sc8 = *(d_schar8_s *)srcPtr;
+
+    rpp_hip_pack_int8(&src_sc8, srcPtr_i8);
+}
+
+__device__ __forceinline__ void rpp_hip_load24_pln3_to_uint24_pln3(uchar *srcPtr, uint increment, d_uint24 *srcPtr_ui24)
+{
+    d_uchar24 src_uc24;
```
rrawther, line 1490: I would discourage the use of this custom datatypes except for few cases where it makes the code.
In this case I would suggest accessing source directly instead of copying the data into src_uc24.
Just use ptrR, pG and pB (cast it to uchar4)

src/include/hip/rpp_hip_common.hpp
```diff
@@ -1429,6 +1466,81 @@ __device__ __forceinline__ void rpp_hip_layouttoggle24_pln3_to_pkd3(T *pixpln3Pt
...
+    srcPtr_ui24->ui4[1] = rpp_hip_pack_uint4(src_uc24.uc4[1]);    // write R04-R07
+    srcPtr_ui24->ui4[2] = rpp_hip_pack_uint4(src_uc24.uc4[2]);    // write G00-G03
+    srcPtr_ui24->ui4[3] = rpp_hip_pack_uint4(src_uc24.uc4[3]);    // write G04-G07
+    srcPtr_ui24->ui4[4] = rpp_hip_pack_uint4(src_uc24.uc4[4]);    // write B00-B03
+    srcPtr_ui24->ui4[5] = rpp_hip_pack_uint4(src_uc24.uc4[5]);    // write B04-B07
+}
+
+__device__ __forceinline__ void rpp_hip_load24_pln3_to_int24_pln3(schar *srcPtr, uint increment, d_int24 *srcPtr_i24)
+{
+    d_schar24_s src_sc24;
```
rrawther, line 1507: please remove this data_type and follow the previous comment

src/modules/cpu/kernel/tensor_sum.hpp
```diff
@@ -0,0 +1,842 @@
...
+                srcPtrRowR += srcDescPtr->strides.hStride;
+                srcPtrRowG += srcDescPtr->strides.hStride;
+                srcPtrRowB += srcDescPtr->strides.hStride;
+            }
+#if __AVX2__
+            rpp_simd_store(rpp_store8_u32_to_u32_avx, sumAvxR, &psumR);
+            rpp_simd_store(rpp_store8_u32_to_u32_avx, sumAvxG, &psumG);
+            rpp_simd_store(rpp_store8_u32_to_u32_avx, sumAvxB, &psumB);
+#endif
+            for(int i = 0; i < 4; i++)
```
rrawther, line 133: unrolling this loop might give better performance

src/modules/cpu/kernel/tensor_sum.hpp
```diff
@@ -0,0 +1,842 @@
...
+                    rpp_simd_load(rpp_load48_u8pln3_to_u32pln3_avx, srcPtrTempR, srcPtrTempG, srcPtrTempB, p);
+                    compute_sum_48_host(p, &psumR, &psumG, &psumB);
+#endif
+                    srcPtrTempR += vectorIncrementPerChannel;
+                    srcPtrTempG += vectorIncrementPerChannel;
+                    srcPtrTempB += vectorIncrementPerChannel;
+                }
+                for (; vectorLoopCount < bufferLength; vectorLoopCount++)
+                {
+                    sumR += (Rpp32u)(*srcPtrTempR);
```
rrawther, line 117: please combine srcPtrTempR++ in this line and get rid of lines 120 to 122

src/modules/cpu/kernel/tensor_sum.hpp
```diff
@@ -0,0 +1,842 @@
...
+                {
+                    sum += (Rpp64f)(*srcPtrTemp);
+                    srcPtrTemp++;
+                }
+                srcPtrRow += srcDescPtr->strides.hStride;
+            }
+#if __AVX2__
+            rpp_simd_store(rpp_store4_f64_to_f64_avx, sumAvx, &psum);
+#endif
+            for(int i = 0; i < 2; i++)
```
rrawther, line 280: please unroll the loop here

src/modules/cpu/kernel/tensor_sum.hpp
```diff
@@ -0,0 +1,842 @@
...
+                    rpp_simd_load(rpp_load24_f32pln3_to_f64pln3_avx, srcPtrTempR, srcPtrTempG, srcPtrTempB, p);
+                    compute_sum_24_host(p, &psumR, &psumG, &psumB);
+#endif
+                    srcPtrTempR += vectorIncrementPerChannel;
+                    srcPtrTempG += vectorIncrementPerChannel;
+                    srcPtrTempB += vectorIncrementPerChannel;
+                }
+                for (; vectorLoopCount < bufferLength; vectorLoopCount++)
+                {
+                    sumR += (Rpp64f)(*srcPtrTempR);
```
rrawther, line 324: see comment above

src/modules/hip/kernel/tensor_sum.hpp
```diff
@@ -0,0 +1,1433 @@
...
+        return;
+    }
+
+    int xAlignedLength = xBufferLength & ~7;                        // alignedLength for vectorized global loads
+    int xDiff = xBufferLength - xAlignedLength;                     // difference between bufferLength and alignedLength
+    uint srcIdx = (id_z * xBufferLength) + id_x;
+
+    d_int8 src_i8;
+    *(d_int8_s *)(&src_i8) = *(d_int8_s *)(srcPtr + srcIdx);
+    if (id_x + 8 > xBufferLength)
```
rrawther, line 73: this code is ambiguous. What is this trying to do?

swetha097: This will help reset invalid values to 0.
Towards right end of the image when width isn't divisible by bufferlength i.e 8 when we could have few junk values, so this segment of code is used to reset those values.

src/modules/cpu/kernel/tensor_sum.hpp
swetha097: Modified code accordingly.

src/modules/cpu/kernel/tensor_sum.hpp
swetha097: Done

swetha097: Modified this in all other locations too.

src/modules/hip/kernel/tensor_sum.hpp
snehaa8: This will help reset invalid values to 0.
Towards right end of the image when width isn't divisible by bufferlength i.e 8, we could have few junk values, so this segment of code is used to reset those values.

src/modules/cpu/kernel/tensor_sum.hpp
snehaa8: Incorporated this change

src/modules/cpu/kernel/tensor_sum.hpp
snehaa8: Incorporated this change

src/modules/cpu/kernel/tensor_sum.hpp
snehaa8: Unrolled the loop here and in all other places.

src/modules/cpu/kernel/tensor_sum.hpp
snehaa8: Addressed this change.

src/modules/cpu/kernel/tensor_sum.hpp
```diff
@@ -0,0 +1,842 @@
...
+                    srcPtrTemp += vectorIncrementPerChannel;
+                }
+                for (; vectorLoopCount < bufferLength; vectorLoopCount++)
+                {
+                    sum += (Rpp32u)(*srcPtrTemp);
+                    srcPtrTemp++;
+                }
+                srcPtrRow += srcDescPtr->strides.hStride;
+            }
+#if __AVX2__
```
rrawther, line 69: what would the code do if AVX2 is not enabled? Applicable to all such

rrawther:
@snehaa8 Please address the review comments
@asalmanp : Can you please review the hip kernels. I think the code need rework

src/modules/cpu/kernel/tensor_sum.hpp
snehaa8: This segment of code is used to store final summation values computed from AVX2 into sumAvx array which can be reduced further with c code, thus enabled this segment only for AVX2.

If AVX2 isn't enabled then usual c summation flow will be followed.

src/modules/cpu/kernel/tensor_sum.hpp
```diff
@@ -0,0 +1,798 @@
...
+#endif
+            for(int i = 0; i < roi.xywhROI.roiHeight; i++)
+            {
+                Rpp8u *srcPtrTemp;
+                srcPtrTemp = srcPtrRow;
+
+                int vectorLoopCount = 0;
+                for (; vectorLoopCount < alignedLength; vectorLoopCount += vectorIncrementPerChannel)
+                {
+#if __AVX2__
```
rrawther, line 55: please move the __AVX2__ flag around the for loop block. After line #52. Otherwise compute won't happen if the flag is not enabled

src/modules/cpu/kernel/tensor_sum.hpp
```diff
@@ -0,0 +1,798 @@
...
+                for (; vectorLoopCount < bufferLength; vectorLoopCount++)
+                {
+                    sum += (Rpp32u)(*srcPtrTemp++);
+                }
+                srcPtrRow += srcDescPtr->strides.hStride;
+            }
+#if __AVX2__
+            rpp_simd_store(rpp_store8_u32_to_u32_avx, sumAvx, &psum);
+#endif
+            sum += (sumAvx[0] + sumAvx[1] + sumAvx[2] + sumAvx[3] + sumAvx[4] + sumAvx[5] + sumAvx[6] + sumAvx[7]);
```
rrawther, line 71: sumAvx will be computed only for __AVX2__.  So this has to go inside the compile flag

rrawther (changes requested):
Please address the review comments

src/modules/cpu/kernel/tensor_sum.hpp
rrawther: I don't see anything changed here. Please commit all your changes based on review comments

rrawther (changes requested):
I think review comments are not addressed for this PR

src/modules/cpu/kernel/tensor_sum.hpp
r-abishek: @rrawther This sum is just the very final addition - for the vector elements in the summed up AVX2 register.
The _mm256_hadd_ps() cannot be used since it only adds adjacent pairs of single-precision f32s. Its better to do the sum[0 through 7] at the end.

src/modules/cpu/kernel/tensor_sum.hpp
r-abishek: The changes for moving __AVX2__ outside the for loop show if the commit is clicked. For some reason, githubUI is not showing the change by clicking the comment. Clicking here shows -> https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/pull/179/commits/b3413d2812c3f7a7d570687b5b8f7e57b90ad516

src/include/hip/rpp_hip_common.hpp
r-abishek: As we discussed all the custom datatypes beyond 24 have all been removed. 

The only reason to retain 24 is the ability to do a one-shot vectorized global memory access - read/write of (RGB pixel * vector of 8 pixels = 24 elements) in a struct. Removing this would need re-evaluation of all loads and stores since they are common to multiple functions. We could run experiments for this, and issue a common PR if load/store funcs need any mods, since this could result in a higher number of global memory accesses causing performance changes.

src/include/hip/rpp_hip_common.hpp
r-abishek: same as above

src/modules/cpu/kernel/tensor_sum.hpp
r-abishek: @rrawther Not sure what's going on with the Github UI. These changes have been addressed though. 

They show up when you open the commit or the file, but not when you click on the comment to verify.
The changes are showing even if we go again to "Files Changed" section or click on "View File".
https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/blob/97f388f1b4df575c70b3b4f01aeb066fbf67964f/src/modules/cpu/kernel/tensor_sum.hpp#L112-L116

src/modules/hip/kernel/tensor_sum.hpp
rrawther: I am not sure if this is the right way to do this having a loop. we need to make the load to load valid pixels instead.
todo:: revisit this code later

rrawther (approved):
@paveltc to run unit tests

src/modules/hip/kernel/tensor_sum.hpp
```diff
@@ -0,0 +1,1433 @@
...
+    int xAlignedLength = roiTensorPtrSrc[id_z].xywhROI.roiWidth & ~7;                           // alignedLength for vectorized global loads
+    int xDiff = roiTensorPtrSrc[id_z].xywhROI.roiWidth - xAlignedLength;                        // difference between roiWidth and alignedLength
+    uint srcIdx = (id_z * srcStridesNCH.x) + ((id_y + roiTensorPtrSrc[id_z].xywhROI.xy.y) * srcStridesNCH.z) + (id_x + roiTensorPtrSrc[id_z].xywhROI.xy.x);
+
+    d_uint24 src_ui24;
+    rpp_hip_load24_pln3_to_uint24_pln3(srcPtr + srcIdx, srcStridesNCH.y, &src_ui24);
+
+    if (id_x + 8 > roiTensorPtrSrc[id_z].xywhROI.roiWidth)                                      // local memory reset of invalid values (from the vectorized global load) to 0
+    {
+        for(int i = xDiff; i < 8; i++)
```
rrawther, line 593: we need to revisit this code segment. I am not sure if this is the right way to prevent invalid load of pixels. Also have a loop here would degrade the overall performance of this kernel

kiritigowda:
@r-abishek can you fix these issues - https://app.codacy.com/gh/GPUOpen-ProfessionalCompute-Libraries/rpp/pullRequest?prid=12196253&bid=36611810

r-abishek:
@kiritigowda 11 Codacy issues resolved. 
Not sure if 2 remaining can be resolved:
d_int24_s is actually used in the code, not sure why codacy points as unused -> https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/blob/ca333f3be7fe8b141604df20cce18bfd72ff2127/src/modules/hip/kernel/tensor_sum.hpp#L247
Similar with d_int8_s -> https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/blob/ca333f3be7fe8b141604df20cce18bfd72ff2127/src/modules/hip/kernel/tensor_sum.hpp#L72

kiritigowda:
> @kiritigowda 11 Codacy issues resolved. Not sure if 2 remaining can be resolved: d_int24_s is actually used in the code, not sure why codacy points as unused ->
> 
> https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/blob/ca333f3be7fe8b141604df20cce18bfd72ff2127/src/modules/hip/kernel/tensor_sum.hpp#L247
> 
> 
> Similar with d_int8_s ->
> https://github.com/GPUOpen-ProfessionalCompute-Libraries/rpp/blob/ca333f3be7fe8b141604df20cce18bfd72ff2127/src/modules/hip/kernel/tensor_sum.hpp#L72

I will mark them as resolved.

r-abishek:
@kiritigowda Also, for this last Codacy issue in this PR, this file has 185 C style castings but it points to only this one casting. 
It's better to handle this later as a separate PR to change all and verify.

paveltc:
@rrawther @kiritigowda @r-abishek Ran the unit tests on both HOST and HIP and everything passes. HIP tests run without issue now.

r-abishek:
@kiritigowda Could we add this PR to the develop to master merge too?

PR #186, opened 2023-08-09
CMakeLists - AMD GPU Target Updates

LakshmiKumar23:
Adding gfx941 to enable rpp on MI300X


CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
 # RPP Default Options
 set(DEFAULT_BUILD_TYPE "Release")
+option(BUILD_WITH_AMD_ADVANCE "Build MIVisionX for advanced AMD GPU Architecture"    OFF)
```
r-abishek, line 51: When would this be ON?

LakshmiKumar23: @r-abishek you will have to turn it on for mi300x. It cannot be made default since it fails with mainline rocm available now

r-abishek: @LakshmiKumar23 okay, so we'll need to modify this to a default YES or a hardware query once there are no issues from rocm, sounds good

r-abishek (approved):

PR #190, opened 2023-08-11
Replace all _lds occurrences to _smem

r-abishek:
@rrawther @asalmanp This PR just replaces all _lds occurrences to Shared in RPP tensor functions.
Once the Gaussian Filter PR is merged, all Shared mem sizes will be changed to the preprocessor define for better readability.

### DIFF
--- a/src/modules/hip/kernel/box_filter.hpp
+++ b/src/modules/hip/kernel/box_filter.hpp
@@ -293,7 +293,7 @@ __global__ void box_filter_3x3_pkd_tensor(T *srcPtr,
     int id_x_i = id_x_o - padLength;
     int id_y_i = id_y_o - padLength;
     d_float24 sum_f24;
-    __shared__ uchar srcShared[48][128];
+    __shared__ uchar src_smem[SMEM_LENGTH_Y_3C][SMEM_LENGTH_X];
 
     int srcIdx = (id_z * srcStridesNH.x) + ((id_y_i + roiTensorPtrSrc[id_z].xywhROI.xy.y) * srcStridesNH.y) + ((id_x_i + roiTensorPtrSrc[id_z].xywhROI.xy.x) * 3);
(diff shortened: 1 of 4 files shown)


src/modules/hip/kernel/box_filter.hpp
```diff
@@ -293,7 +293,7 @@ __global__ void box_filter_3x3_pkd_tensor(T *srcPtr,
     int id_x_i = id_x_o - padLength;
     int id_y_i = id_y_o - padLength;
     d_float24 sum_f24;
-    __shared__ uchar src_lds[48][128];
+    __shared__ uchar srcShared[48][128];
```
asalmanp, line 296: I prefer the current name src_lds. Also, based on the google coding standard, we should avoid the CamelCase for variable names. please see [here](https://google.github.io/styleguide/cppguide.html#Constant_Names:~:text=enum%20class%20UrlTableError%20%7B%20...-,Variable%20Names,-The%20names%20of). The only thing that I would suggest is to replace the 48, and 128 with the defines like the Gaussian filter PR.

asalmanp (changes requested):

src/modules/hip/kernel/box_filter.hpp
rrawther: Hi @asalmanp :  When attending HIP training, they mentioned LDS is AMD's internal name so it is better not to use that name. Outside people doesn't really know what lds means. May I suggest src_shared[] instead?

rrawther (changes requested):
Please change the case. Please don't mix camel and snake case

kiritigowda:
@r-abishek any update on this?

rrawther (approved):
@paveltc to run tests

asalmanp (approved):

r-abishek:
@kiritigowda Codacy shows X, but it shows me no issues if I click on 'Details'

paveltc:
@rrawther @kiritigowda This PR passes the unit tests.

PR #251, opened 2023-10-31
Readme & changelog updates

LisaDelaney:


CHANGELOG.md
```diff
@@ -1,341 +1,379 @@
...
-
-#### Fixed
-* 
-
-### Tested Configurations
 * Linux distribution
-  + Ubuntu - `20.04` / `22.04`
-  + CentOS - `7`
-  + RHEL - `8`/`9`
+  * Ubuntu - `20.04` / `22.04`
```
kiritigowda, line 14: Is this nested bullet format required for visualization? If not, can we leave the current format?

LisaDelaney: Sorry, I'm not sure I understand your ask. The current version also has nested bullets, as the OSs listed fall under Linux distributions.

saadrahim: I also think we need to remove the supported OS list here and apply ROCm support only

kiritigowda: @saadrahim some libraries only support a subset of ROCm supported OS. I think it is good to just list what the library supports.

kiritigowda: @LisaDelaney thanks for making all the changes. Our team uses the VS code markdown formatter and I think it defaults to using `*`/`+`/`-` for nested bullets. We can make that change to config file to stay consistent.

LisaDelaney: @kiritigowda I prefer using different style bullets for different levels...sadly, our linter doesn't approve. :/

kiritigowda (approved):

PR #264, opened 2023-11-21
Kg/package

kiritigowda:
* Package Updates
* Test Package


r-abishek (commented):
I see you've changed runPackageCommand and the cmake to separate the different rpm or deb packages. This might be useful for future cleaner support - when we decide to run all OS variants in CI for all backends too?

Issue #24, opened 2020-04-29
RPP - Ubuntu-16.04 Support


LakshmiKumar23:
Build fails on Ubuntu-16.04:
`
[  2%] Building CXX object addkernels/CMakeFiles/addkernels.dir/include_inliner.cpp.o
[  4%] Building CXX object addkernels/CMakeFiles/addkernels.dir/addkernels.cpp.o
In file included from /root/mivisionx-deps/rpp/addkernels/addkernels.cpp:26:
In file included from /root/mivisionx-deps/rpp/addkernels/include_inliner.hpp:29:
/root/mivisionx-deps/rpp/addkernels/source_file_desc.hpp:37:10: error: no type named 'shared_ptr' in namespace 'std'
    std::shared_ptr<SourceFileDesc> included_from;
    ~~~~~^
/root/mivisionx-deps/rpp/addkernels/source_file_desc.hpp:37:20: error: expected member name or ';' after declaration specifiers
    std::shared_ptr<SourceFileDesc> included_from;
    ~~~~~~~~~~~~~~~^
/root/mivisionx-deps/rpp/addkernels/source_file_desc.hpp:39:51: error: no type named 'shared_ptr' in namespace 'std'
    SourceFileDesc(const std::string& path_, std::shared_ptr<SourceFileDesc> from, int line)
                                             ~~~~~^
/root/mivisionx-deps/rpp/addkernels/source_file_desc.hpp:39:61: error: expected ')'
    SourceFileDesc(const std::string& path_, std::shared_ptr<SourceFileDesc> from, int line)
                                                            ^
/root/mivisionx-deps/rpp/addkernels/source_file_desc.hpp:39:19: note: to match this '('
    SourceFileDesc(const std::string& path_, std::shared_ptr<SourceFileDesc> from, int line)
                  ^
/root/mivisionx-deps/rpp/addkernels/source_file_desc.hpp:40:38: error: use of undeclared identifier 'line'
        : path(path_), included_line(line), included_from(from)
                                     ^
/root/mivisionx-deps/rpp/addkernels/source_file_desc.hpp:40:59: error: use of undeclared identifier 'from'
        : path(path_), included_line(line), included_from(from)
                                                          ^
In file included from /root/mivisionx-deps/rpp/addkernels/addkernels.cpp:26:
/root/mivisionx-deps/rpp/addkernels/include_inliner.hpp:34:7: error: exception specification of overriding function is more lax than base version
[...]

PR #271, opened 2023-12-13
RPP Color Temperature on HOST and HIP

r-abishek:
- Adds image color temperature adjustment support on HOST and HIP for u8/f16/f32/i8
- Adds relevant QA/unit/perf tests


src/include/cpu/rpp_cpu_common.hpp
```diff
@@ -3028,6 +3028,22 @@ inline void compute_color_cast_12_host(__m128 *p, __m128 pMul, __m128 *pAdd)
     p[2] = _mm_fmadd_ps(_mm_sub_ps(p[2], pAdd[2]), pMul, pAdd[2]);    // color_cast adjustment Rs
 }
 
+inline void compute_color_temperature_48_host(__m256 *p, __m256 pAdj)
```
rrawther, line 3031: please rename the variable name to snake_case

src/modules/cpu/kernel/color_temperature.hpp
```diff
@@ -0,0 +1,984 @@
...
+
+        Rpp32f adjustmentValue = adjustmentValueTensor[batchCount];
+
+        Rpp8u *srcPtrImage, *dstPtrImage;
+        srcPtrImage = srcPtr + batchCount * srcDescPtr->strides.nStride;
+        dstPtrImage = dstPtr + batchCount * dstDescPtr->strides.nStride;
+
+        Rpp32u bufferLength = roi.xywhROI.roiWidth * layoutParams.bufferMultiplier;
+
+        __m256 pAdj = _mm256_set1_ps(adjustmentValue);
```
rrawther, line 32: please combine line #24 into this and remove line #24

src/modules/hip/kernel/color_temperature.hpp
```diff
@@ -0,0 +1,233 @@
+#include <hip/hip_runtime.h>
+#include "rpp_hip_common.hpp"
+
+__device__ void color_temperature_hip_compute(uchar *srcPtr, d_float24 *pix_f24, float4 *adjustmentValue_f4)
+{
```
rrawther, line 5: why is the srcPtr passed to this function?

utilities/test_suite/REFERENCE_OUTPUT/color_temperature/color_temperature_u8_Tensor_PKD3.csv
rrawther, line 1: Please combine both csv files

rrawther (changes requested):
Please address review comments

src/modules/cpu/kernel/color_temperature.hpp
r-abishek: We will need the adjustmentValue parameter in 8 other locations below though, which are all inside loops within one large omp loop. So it might be better to create a local variable for that?

src/modules/hip/kernel/color_temperature.hpp
r-abishek: We had a discussion long ago on this, I think we should discuss these on a separate short call.

utilities/test_suite/REFERENCE_OUTPUT/color_temperature/color_temperature_u8_Tensor_PKD3.csv
r-abishek: @rrawther For the ~31*3 = 93 subvariants we already have in the new test_suite, we have PLN3/PKD3/PLN1 CSVs, but we could possibly combine them if each file size doesn't grow too large, and if accessing the required variant's output from the file is easy enough - hopefully not too confusing for changes later.
Each CSV currently contains the 50x50, 100x100 and 150x150 = ~ 35,000* 3 channels = ~ 100,000 elements.

Combining the PKD3 and PLN3 outputs would possibly have 200,000 elements in that CSV, but we'll need appropriate test suite logic to jump to the right location and read only necessary lines from the file.
@snehaa8 We can check what particular lines need to change in the test suite possibly?

src/include/cpu/rpp_cpu_common.hpp
snehaa8: Rajy will get back on this.

src/modules/cpu/kernel/color_temperature.hpp
snehaa8: Sure Rajy I will modify this.

src/modules/hip/kernel/color_temperature.hpp
snehaa8: Sure Rajy we will check performance with Templated version as discussed.

utilities/test_suite/REFERENCE_OUTPUT/color_temperature/color_temperature_u8_Tensor_PKD3.csv
snehaa8: For now I will combine them into one csv then we will work on disabling AVX and SSE flags on compile time in cmake.

src/include/cpu/rpp_cpu_common.hpp
snehaa8: Performance numbers are similar to using templated function for each datatype with minor differences due to noise. Modified them accordingly.

src/modules/hip/kernel/color_temperature.hpp
snehaa8: Performance numbers are similar to using templated function for each datatype with minor differences due to noise. Modified them accordingly.

utilities/test_suite/REFERENCE_OUTPUT/color_temperature/color_temperature_u8_Tensor_PKD3.csv
snehaa8: Used .bin file as a temporary change.

src/include/cpu/rpp_cpu_common.hpp
r-abishek: Rajy, this is actually consistent with all other API. Camel case for variables and snake case for RPP API.

rrawther (approved):

PR #276, opened 2023-12-20
Testing - change output folder image writes, data dumps, performance result writes to CWD

r-abishek:
@kiritigowda This is a minor PR that changes output writes from testing to CWD.

### DIFF
--- a/utilities/test_suite/HIP/runTests.py
+++ b/utilities/test_suite/HIP/runTests.py
@@ -33,7 +33,7 @@ inFilePath1 = scriptPath + "/../TEST_IMAGES/three_images_mixed_src1"
 inFilePath2 = scriptPath + "/../TEST_IMAGES/three_images_mixed_src2"
 ricapInFilePath = scriptPath + "/../TEST_IMAGES/three_images_150x150_src1"
 qaInputFile = scriptPath + "/../TEST_IMAGES/three_images_mixed_src1"
-outFolderPath = scriptPath + "/../../../build"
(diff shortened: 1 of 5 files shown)


utilities/test_suite/HIP/runTests.py
```diff
@@ -33,6 +33,7 @@
 inFilePath2 = scriptPath + "/../TEST_IMAGES/three_images_mixed_src2"
 ricapInFilePath = scriptPath + "/../TEST_IMAGES/three_images_150x150_src1"
 qaInputFile = scriptPath + "/../TEST_IMAGES/three_images_mixed_src1"
+outFolderPath = scriptPath + "/../../../build"
```
r-abishek, line 36: @kiritigowda The only assumption I make here is that the folder name is always "build" directly under "rpp". Is that okay?

r-abishek: Fixed. Changed from build folder to CWD.

kiritigowda:
@r-abishek need to use std filesystem for get_current_dir_name as its supported everywhere.

#include <filesystem>
#define get_current_dir_name std::filesystem::current_path

kiritigowda (approved):

kiritigowda:
Tensor_host/hip - needs to be built in current path

r-abishek:
@kiritigowda Could you merge this one?
Issuing a separate PR for moving the build folder of Tensor_hip/host to CWD. Don't want to mix with output folder dump changes

PR #285, opened 2024-01-18
RPP Voxel Flip on HIP and HOST

r-abishek:
- Adds voxel flip on HIP and HOST
- Adds necessary unit/perf/QA tests


src/modules/cpu/kernel/flip_voxel.hpp
```diff
@@ -0,0 +1,418 @@
+/*
+Copyright (c) 2019 - 2023 Advanced Micro Devices, Inc. All rights reserved.
```
rrawther, line 2: change to 2024

src/modules/cpu/kernel/flip_voxel.hpp
```diff
@@ -0,0 +1,418 @@
...
+
+#include "rppdefs.h"
+#include "rpp_cpu_simd.hpp"
+#include "rpp_cpu_common.hpp"
+
+RppStatus flip_voxel_f32_f32_host_tensor(Rpp32f *srcPtr,
+                                         RpptGenericDescPtr srcGenericDescPtr,
+                                         Rpp32f *dstPtr,
+                                         RpptGenericDescPtr dstGenericDescPtr,
+                                         Rpp32u *horizontalTensor,
```
rrawther, line 31: what are these horizontal, vertical and depth tensors for?

src/modules/hip/kernel/flip_voxel.hpp
```diff
@@ -0,0 +1,188 @@
...
+                                  int channels,
+                                  uint3 mirrorXYZ,
+                                  RpptROI3DPtr roiGenericPtrSrc,
+                                  int batchIndex)
+{
+    int id_x = (hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x) * 8;        // W - inner most dim vectorized
+    int id_y = hipBlockIdx_y * hipBlockDim_y + hipThreadIdx_y;              // H - second to inner
+    int id_z = hipBlockIdx_z * hipBlockDim_z + hipThreadIdx_z;              // D - outer most dim
+
+    if ((id_z >= roiGenericPtrSrc->xyzwhdROI.roiDepth) || (id_y >= roiGenericPtrSrc->xyzwhdROI.roiHeight) || (id_x >= roiGenericPtrSrc->xyzwhdROI.roiWidth))
```
rrawther, line 18: we are assuming the roi always start from 0 for all dims. Is there a case otherwise?

src/modules/hip/kernel/flip_voxel.hpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    int dstIdx = (id_z * dstStridesCDH.y) + (id_y * dstStridesCDH.z) + id_x;
+    int zFactor = ((mirrorXYZ.z) ? (roiGenericPtrSrc->xyzwhdROI.roiDepth - 1 - id_z) : (id_z + roiGenericPtrSrc->xyzwhdROI.xyz.z)) * srcStridesCDH.y;
+    int yFactor = ((mirrorXYZ.y) ? (roiGenericPtrSrc->xyzwhdROI.roiHeight - 1 - id_y) : (id_y + roiGenericPtrSrc->xyzwhdROI.xyz.y)) * srcStridesCDH.z;
+    int xFactor = id_x + roiGenericPtrSrc->xyzwhdROI.xyz.x;
+    if (mirrorXYZ.x)
+    {
+        // To handle the case when trying to load from invalid memory location when width is not a multiple of 8
+        if((batchIndex == 0) && (id_x + 8 > roiGenericPtrSrc->xyzwhdROI.roiWidth))
```
rrawther, line 30: why the kernel need to differentiate batchIndex? I think we need to do a code review for this PR

rrawther (commented):
Let us have a code review to go over this PR

src/modules/cpu/kernel/flip_voxel.hpp
sampath1117: Hi Rajy,
These are parameters that are used for determining the type of flip needed. They can be 0/1
horizontalTensor - if set to 1, data is flipped across horizontal direction
verticalTensor - if set to 1, data is flipped across vertical direction
depthTensor - if set to 1, data is flipped across depth direction

src/modules/hip/kernel/flip_voxel.hpp
sampath1117: Hi Rajy,
we are not assuming ROI to start from 0 for all dims. we are using roiWidth, roiHeight, roiDepth as the boundary conditions, which is the size of output

we are using roi, x, y, z co-ordinates here for loading the corresponding input location
https://github.com/r-abishek/rpp/blob/ar/flip_voxel/src/modules/hip/kernel/flip_voxel.hpp#L24

src/modules/hip/kernel/flip_voxel.hpp
sampath1117: For voxel kernels, we are launching each sample in a batch in a separate kernel
https://github.com/r-abishek/rpp/blob/ar/flip_voxel/src/modules/hip/kernel/flip_voxel.hpp#L147

This differentiation is needed when horizontal Flag is set to 1
in the kernel, we are loading as 8 pixel load and store and we are trying to load input in reverse manner when  horizontal Flag is set to 1

for output location 0, we are trying load 8 values from input (width - 1 - 8)th location. When the output width is a not a multiple of 8. The above formulae will result in negative source location when output location is  (width - 1 - 8). So to handle that case, we have added a separate condition for batchIndex = 0

src/modules/cpu/kernel/flip_voxel.hpp
rrawther: It is better to have a single tensor of size [bs][3] for representing these. In other words please combine all the flip flags into a single tensor

rrawther (changes requested):

src/modules/cpu/kernel/flip_voxel.hpp
shobana-mcw: Hi Rajy,

The DALI API uses 3 different tensors for horizontal, vertical and depth arguments. So we are following the same in rocAL python and C++ API too.

rrawther (commented):
@r-abishek  Why there are 285 file changes in this? 
@kiritigowda : Probably you know about these changes more. Please review in that case

rrawther:
The PR says WIP. Is this still in development?

r-abishek:
> @r-abishek Why there are 285 file changes in this? @kiritigowda : Probably you know about these changes more. Please review in that case

@rrawther This PR is streamlined now

rrawther (approved):
@paveltc to test

kiritigowda:
@r-abishek failing CI

r-abishek:
> @r-abishek failing CI

@kiritigowda It seems to be failing in the docker build itself

For RHEL it says:
Failed to run image 'compute-rocm-dkms-no-npi-hipclang-13402/rhel9gfx1101/math-ci:latest'. Error: docker: Error response from daemon: error gathering device information while adding custom device "/dev/kfd": no such file or directory

And for Ubuntu:
mkdir /var/lib/docker/containers/166b8658462a57a2c20eb92e78021e5bd7c7c14f95105feb81e26c404fd59fab: no such file or directory

kiritigowda:
@r-abishek  - I did a rerun of the CI and seeing a pass - http://math-ci.amd.com/blue/organizations/jenkins/main%2Fprecheckin%2Frpp/detail/PR-285/14/pipeline

Can you verify if we got all the tests covered? I can merge this.

Issue #291, opened 2024-01-19
[Issue]:  GFX942+ Compile Failure


kiritigowda:
### Problem Description

Building with `-D BUILD_WITH_AMD_ADVANCE=ON`  

```
[ 70%] Building CXX object src/modules/CMakeFiles/modules.dir/hip/kernel/hue.cpp.o
lld: error: undefined hidden symbol: llvm.amdgcn.atomic.inc.i32.p0i32
>>> referenced by /tmp/histogram-gfx941-aebaf1.o:(partial_histogram_pln)
>>> referenced by /tmp/histogram-gfx941-aebaf1.o:(partial_histogram_pln)
>>> referenced by /tmp/histogram-gfx941-aebaf1.o:(partial_histogram_pkd)
>>> referenced 5 more times
clang++: error: amdgcn-link command failed with exit code 1 (use -v to see invocation)
make[2]: *** [src/modules/CMakeFiles/modules.dir/build.make:1302: src/modules/CMakeFiles/modules.dir/hip/kernel/histogram.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:918: src/modules/CMakeFiles/modules.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

```shell
[ 72%] Building CXX object src/modules/CMakeFiles/modules.dir/hip/kernel/integral.cpp.o
lld: error: undefined hidden symbol: llvm.amdgcn.atomic.inc.i32.p0i32
>>> referenced by /tmp/histogram-gfx941-d9e6b1.o:(partial_histogram_pln)
>>> referenced by /tmp/histogram-gfx941-d9e6b1.o:(partial_histogram_pln)
>>> referenced by /tmp/histogram-gfx941-d9e6b1.o:(partial_histogram_pkd)
>>> referenced 5 more times
clang++: error: amdgcn-link command failed with exit code 1 (use -v to see invocation)
make[2]: *** [src/modules/CMakeFiles/modules.dir/build.make:1302: src/modules/CMakeFiles/modules.dir/hip/kernel/histogram.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:918: src/modules/CMakeFiles/modules.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

### Operating System

ALL

### CPU

ALL

### GPU

Other

### Other

ALL

[...]

PR #295, opened 2024-01-29
Build fix for ROCm 6.0.0 gfx942

r-abishek:
@kiritigowda Fix for #291

### DIFF
--- a/src/modules/hip/kernel/histogram.cpp
+++ b/src/modules/hip/kernel/histogram.cpp
@@ -39,9 +39,9 @@ void partial_histogram_pln( unsigned char *input,
         unsigned char pixelR = input[pixId];
         unsigned char pixelG = input[pixId + width * height];
         unsigned char pixelB = input[pixId + 2 * width * height];
-        atomicInc(&tmp_histogram[pixelR], 1);
-        atomicInc(&tmp_histogram[pixelG], 1);
-        atomicInc(&tmp_histogram[pixelB], 1);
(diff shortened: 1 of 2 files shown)


kiritigowda (approved):

src/modules/hip/kernel/histogram.cpp
```diff
@@ -39,9 +39,9 @@ void partial_histogram_pln( unsigned char *input,
         unsigned char pixelR = input[pixId];
         unsigned char pixelG = input[pixId + width * height];
         unsigned char pixelB = input[pixId + 2 * width * height];
-        atomicInc(&tmp_histogram[pixelR], 1);
-        atomicInc(&tmp_histogram[pixelG], 1);
-        atomicInc(&tmp_histogram[pixelB], 1);
+        atomicAdd(&tmp_histogram[pixelR], 1);
```
rrawther, line 42: Is atomicAdd good option. I thought we can't use any atomics

rrawther (commented):
Added a comment

src/modules/hip/kernel/histogram.cpp
r-abishek: @rrawther Ubuntu build seems to pass with atomicAdd. Ideally, both should work as they did on openCL/cuda/hip, but for some reason only atomicInc() fails on Rocm6.0.0 HIP

kiritigowda (approved):

PR #296, opened 2024-01-29
RPP Tensor Audio Support - Down Mixing

r-abishek:
- Adds support for Audio functionality - Down Mixing
- Adds relevant QA and performance tests


src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -2438,6 +2438,29 @@ static inline __m128 log_ps(__m128 x)
     return x;
 }
 
+inline Rpp32f rpp_hsum_ps(__m128 x)
+{
+    __m128 shuf = _mm_movehdup_ps(x);        // broadcast elements 3,1 to 2,0
+    __m128 sums = _mm_add_ps(x, shuf);
```
rrawther, line 2444: why aren't we using built-in hadd instructions?

src/modules/cpu/kernel/down_mixing.hpp
```diff
@@ -0,0 +1,123 @@
...
+*/
+
+#include "rppdefs.h"
+#include <omp.h>
+
+RppStatus down_mixing_host_tensor(Rpp32f *srcPtr,
+                                  RpptDescPtr srcDescPtr,
+                                  Rpp32f *dstPtr,
+                                  RpptDescPtr dstDescPtr,
+                                  Rpp32s *srcLengthTensor,
```
rrawther, line 32: It is better to combine both channels and Length into a single tensor

rrawther (changes requested):

src/include/cpu/rpp_cpu_simd.hpp
```diff
@@ -2438,6 +2438,29 @@ static inline __m128 log_ps(__m128 x)
     return x;
 }
 
+inline Rpp32f rpp_hsum_ps(__m128 x)
+{
```
LakshmiKumar23, line 2442: @r-abishek Are we not following Google style coding for RPP? With all opening { on the same line as function name

src/include/cpu/rpp_cpu_simd.hpp
sampath1117: Hi Rajy,
With the usage of built-in  _mm_hadd_ps, the performance was slower,  so we used the above alternate method for doing sum of all values in __m128 register. The above approach is nearly 20% faster than the approach with built-in instruction

src/include/cpu/rpp_cpu_simd.hpp
r-abishek: Once the BatchPD->Tensor migration is complete and we are in a position to phase out BatchPD code, there will be a lot less code to deal with, and a single PR to format the whole repo could be done to adhere to google std.
For now, there's easily a 90+% majority of code in RPP in this format, so retaining this

src/include/cpu/rpp_cpu_simd.hpp
rrawther: Can you please make a sample code with both ways. We need to run it on different CPU versions

sampath1117: H Rajy,
Adding a snippet of sample for both variations for your reference
image

In Intel intrinsic guide, _mm_hadd_ps latency is available only for Skylake 
On Skylake, Built-in variant takes a total latency of 19 cycles whereas custom variant takes 14 cycles

src/modules/rppt_tensor_audio_augmentations.cpp
```diff
@@ -126,3 +126,31 @@ RppStatus rppt_pre_emphasis_filter_host(RppPtr_t srcPtr,
...
+                                dstDescPtr,
+                                srcDimsTensor,
+                                normalizeWeights,
+                                rpp::deref(rppHandle));
+
+        return RPP_SUCCESS;
+    }
+    else
+    {
+        return RPP_ERROR_NOT_IMPLEMENTED;
```
rrawther, line 154: Don't we need to have FP16 as well?

rrawther (commented):
Added comment

src/modules/rppt_tensor_audio_augmentations.cpp
r-abishek: All 9 audio functions are currently in F32. Especially with functions using Decibels, small precision changes are affecting final outputs or training accuracy. 
We can possibly do a F16-study on porting all these once we have all 9 in F32 in rpp ToT.

r-abishek:
@rrawther This PR has tests passing and comments addressed. Could be reviewed

rrawther (approved):

PR #311, opened 2024-02-21
RPP Vignette Tensor on HOST and HIP

r-abishek:
- Adds support for Vignette on HOST and HIP
- Adds support for NCHW/NHWC variants with toggle for U8/F16/F32/I8 depths
- Adds support for relevant PASS/FAIL QA tests, Unit tests with image dumps and performance tests.


LakshmiKumar23 (approved):
Tested with test_case 46 for HOST and HIP

PR #316, opened 2024-03-04
Png update

randyh62:
Convert JPG files to PNG format
Remove JP files from docs/data/doxygenInputs and docs/data/doxygenOutputs folder
update header files to point to PNG image files


kiritigowda:
@r-abishek can you review this PR and check if the standalone doc generation works?

r-abishek:
@kiritigowda Verified Local machine Doxygen build still shows all images as below after the jpg->png conversion
image

r-abishek (approved):

docs/doxygen/Doxyfile
```diff
@@ -1117,8 +1117,8 @@ EXAMPLE_RECURSIVE      = NO
 # that contain images that are to be included in the documentation (see the
 # \image command).
 
-IMAGE_PATH             = ../../data/doxygenInputs \
-                         ../../data/doxygenOutputs
+IMAGE_PATH             = ../data/doxygenInputs \
```
r-abishek, line 1120: Interesting that the path change didn't affect anything on local build. Images show before and now.

randyh62: It was interesting that it built with the ../../ path when you consider the Doxyfile is in docs/doxygen and the images are in docs/data. It did not build correctly on my system, which is why I changed it.

PR #357, opened 2024-05-17
RPP Glitch on HOST and HIP

r-abishek:
- Adds support for glitch augmentation on host and hip
- Adds relevant unit/performance/QA tests


src/modules/cpu/kernel/glitch.hpp
```diff
@@ -0,0 +1,692 @@
...
+*/
+
+#include "rppdefs.h"
+#include "rpp_cpu_simd.hpp"
+#include "rpp_cpu_common.hpp"
+
+inline void compute_src_loc(int row , int col, Rpp32s *locArray, RpptDescPtr srcDescPtr, RpptChannelOffsets *rgbOffsets, RpptROI roi, int batchCount, int channelValue)
+{
+    int xR, yR, xG, yG, xB, yB;
+    xR = col + rgbOffsets[batchCount].r.x;
```
rrawther, line 32: can we make sure the rgbOffsets[]... are within valid boundary thereby avoiding the boundary checks below

src/modules/cpu/kernel/glitch.hpp
```diff
@@ -0,0 +1,692 @@
...
+
+                dstPtrRow += dstDescPtr->strides.hStride;
+            }
+        }
+        else if ((srcDescPtr->c == 3) && (srcDescPtr->layout == RpptLayout::NCHW) && (dstDescPtr->layout == RpptLayout::NHWC))
+        {
+            Rpp32f *dstPtrRow;
+            dstPtrRow = dstPtrChannel;
+            Rpp32u vectorIncrement = 8;
+            Rpp32u alignedLength = (static_cast<int>(roi.xywhROI.roiWidth * 0.75)) & ~7;   // Considering 75% of the ROI width to avoid loading out-of-bounds locations.
```
rrawther, line 306: as we discussed, this can cause some areas visibily showing without glitching

rrawther (changes requested):
Added some comments

src/modules/cpu/kernel/glitch.hpp
r-abishek: Discussed this, and the only solution here without checking bounds might be to just trust the user/caller that they will give the values within bounds, which may not necessarily be true. 
Agreed to have the checks on the hip side at least for now.

src/modules/cpu/kernel/glitch.hpp
r-abishek: Done, removed the 0.75, changed to nearest multiple of alignedLength

src/modules/hip/kernel/glitch.hpp
```diff
@@ -0,0 +1,278 @@
...
+
+template <typename T>
+__device__ __forceinline__ void rpp_hip_load1_glitch(T *srcPtr, uint2 srcStrideCH, float &locSrcX, float &locSrcY, float *dst, int channels)
+{
+    int srcIdx = locSrcY * srcStrideCH.y + locSrcX * srcStrideCH.x + channels;
+    rpp_hip_interpolate1_nearest_neighbor_load_pln1(srcPtr + srcIdx, dst);
+}
+
+template <typename T>
+__device__ __forceinline__ void rpp_hip_load8_glitch(T *srcPtr, uint2 srcStrideCH, d_float8 *srcX_f8, d_float8 *srcY_f8, d_float8 *dst_f8, int channels)
```
rrawther, line 12: @AryanSalmanpour: Can you please review this kernel?

rrawther (approved):
@AryanSalmanpour: to review hip kernel before merging

src/modules/hip/kernel/glitch.hpp
r-abishek: @rrawther Did the review on the call for this last week with @AryanSalmanpour
Added some comments in HIP kernels as per suggestions - https://github.com/ROCm/rpp/pull/357/commits/7d5855b7e36fb7e91c3ccc81fbd98d2c415c915d

AryanSalmanpour (approved):

PR #372, opened 2024-06-10
RPP Log Tensor on HOST and HIP

r-abishek:
- Adds support for log tensor on host and hip for u8/f16/f32/i8 bit depths
- Adds relevant unit/qa/performance tests


utilities/test_suite/HIP/Tensor_misc_hip.cpp
```diff
@@ -31,7 +31,7 @@ int main(int argc, char **argv)
     if (argc < MIN_ARG_COUNT)
     {
         printf("\nImproper Usage! Needs all arguments!\n");
-        printf("\nUsage: ./Tensor_misc_hip <case number = 0:0> <test type 0/1> <toggle 0/1> <number of dimensions> <batch size> <num runs> <dst path> <script path>\n");
```
rrawther, line 34: I see mix of naming convensions here. Let us discuss it sometime over a call

rrawther (approved):

PR #373, opened 2024-06-10
Bugfix for Erase variant

r-abishek:
Adds a minor bugfix for erase HIP PLN3-PKD3 variant correcting kernel call and templating for multi-bit depth support


rrawther (approved):
Please merge with top if needed

PR #378, opened 2024-06-12
Fix to exclude linking FFTS and add AUDIO_SUPPORT flag

r-abishek:
- Adds an AUDIO_SUPPORT definition in RPP
- Excludes linking and audio support when RPP_AUDIO_SUPPORT=OFF
- Similar to the rpp_backend.h that tells what backend the user installed, introduces a rpp_audio_augmentations_support.h to tell if RPP was installed with or without audio support by the user, and based on that run CI / test suites avoiding any failures.


CMakeLists.txt
```diff
@@ -59,6 +59,16 @@ endif(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 # RPP Default Options
 set(DEFAULT_BUILD_TYPE "Release")
 
+# RPP Audio support - (default = ON, support currently on Ubuntu - user to set to OFF otherwise)
+if(NOT DEFINED RPP_AUDIO_SUPPORT)
```
LakshmiKumar23, line 63: is this an option for the user?
please give it as option in cmake then similar to this https://github.com/ROCm/MIVisionX/blob/fcb9876138d9bb7389785d4e958266d84627b824/CMakeLists.txt#L51

r-abishek: yeah, modified this to use option() and added some more support

r-abishek:
Output on MacOS for `cmake -D BACKEND=CPU -D RPP_AUDIO_SUPPORT=OFF ..`
Screen Shot 2024-06-13 at 8 33 27 PM

r-abishek:
Output on MacOS for `cmake -D BACKEND=CPU ..`
Screen Shot 2024-06-13 at 8 36 35 PM

kiritigowda:
@r-abishek -- needs to resolve conflicts

r-abishek:
@kiritigowda resolved

kiritigowda (approved):

PR #382, opened 2024-06-17
RPP Lens Correction on HOST and HIP

r-abishek:
- Adds support for Lens Correction on HOST and HIP with bit-depth support for u8/f16/f32/i8.
- Adds relevant unit/performance/qa tests.


include/rppt_tensor_geometric_augmentations.h
```diff
@@ -634,6 +634,58 @@ RppStatus rppt_remap_host(RppPtr_t srcPtr, RpptDescPtr srcDescPtr, RppPtr_t dstP
 RppStatus rppt_remap_gpu(RppPtr_t srcPtr, RpptDescPtr srcDescPtr, RppPtr_t dstPtr, RpptDescPtr dstDescPtr, Rpp32f *rowRemapTable, Rpp32f *colRemapTable, RpptDescPtr tableDescPtr, RpptInterpolationType interpolationType, RpptROIPtr roiTensorPtrSrc, RpptRoiType roiType, rppHandle_t rppHandle);
 #endif // GPU_SUPPORT
 
+/*! \brief Lens correction transformation on HOST backend for a NCHW/NHWC layout tensor
+ * \details Performs lens correction transforms on an image to compensate barrel lens distortion of RGB(3 channel) / greyscale(1 channel) images with an NHWC/NCHW tensor layout.<br>
+ * - srcPtr depth ranges - Rpp8u (0 to 255), Rpp16f (0 to 1), Rpp32f (0 to 1), Rpp8s (-128 to 127).
+ * - dstPtr depth ranges - Will be same depth as srcPtr.
+ * \image html lens_img640x480.png Sample Input
```
rrawther, line 641: Do we support other image sizes? Please modify the description accordingly. Otherwise it looks like we use hardcoded image path

src/modules/cpu/kernel/lens_correction.hpp
```diff
@@ -0,0 +1,177 @@
...
+#include "rppdefs.h"
+#include "rpp_cpu_simd.hpp"
+#include "rpp_cpu_common.hpp"
+#include <omp.h>
+
+// Compute Inverse matrix (3x3)
+inline void get_inverse(float *mat, float *invMat)
+{
+    float det = mat[0] * (mat[4] * mat[8] - mat[7] * mat[5]) - mat[1] * (mat[3] * mat[8] - mat[5] * mat[6]) + mat[2] * (mat[3] * mat[7] - mat[4] * mat[6]);
+    if(det != 0)
```
rrawther, line 34: what will be invMat for else case?

src/modules/hip/kernel/lens_correction.hpp
```diff
@@ -0,0 +1,176 @@
...
+                                          rpp::Handle& handle)
+{
+    if (roiType == RpptRoiType::LTRB)
+        hip_exec_roi_converison_ltrb_to_xywh(roiTensorPtrSrc, handle);
+
+    int globalThreads_x = (dstDescPtr->w + 7) >> 3;
+    int globalThreads_y = dstDescPtr->h;
+    int globalThreads_z = dstDescPtr->n;
+
+    float *inverseMatrix = handle.GetInitHandle()->mem.mgpu.scratchBufferHip.floatmem;
```
rrawther, line 154: Is this matrix constant across batch? If so it is beneficial to pass precomputed inverse matrix to the kernel instead

rrawther (changes requested):

include/rppt_tensor_geometric_augmentations.h
r-abishek: This one is just a pointer to a file so that doxygen shows a sample input / output as we have done for other functions too

src/modules/cpu/kernel/lens_correction.hpp
r-abishek: @rrawther Inverse of a matrix is not defined when its determinant is 0. 
So in those scenarios, it follows that we can't run this augmentation. It could also have different results for different images in the batch, so we can't return a single enum for failure.
A black image could be returned for that image in the batch, with a note for this in the header docs stating it will return a black image if the user provided camera matrix has a 0 determinant.

src/modules/hip/kernel/lens_correction.hpp
r-abishek: Its a user-provided 9-element matrix that could vary for each image in the batch

rrawther (approved):

PR #331, opened 2024-03-27
Docs - Missing input and output images for Doxygen

r-abishek:
@kiritigowda This PR adds missing input/output images for Doxygen docs.

PR #385, opened 2024-06-19
Docs - Missing input and output images for Doxygen (#331)

randyh62:
* added missing outputs for image augmentations

modified the description with correct output names

* added gif for voxel input and outputs

* modified the output images for water, resize_crop_mirror and resize_mirror_normalize

---------
This never made it into 6.1.2


r-abishek (approved):
These should fix the missing images in the public docs. #331 added some of them in ToT - master.

PR #395, opened 2024-06-28
RPP Audio Support HIP - Non silent region

r-abishek:
- Adds RPP audio support for Non Silent Region on HIP backend.
- Adds relevant qa tests and performance tests.


src/modules/hip/kernel/non_silent_region_detection.hpp
```diff
@@ -0,0 +1,419 @@
...
+                                                      Rpp32s *detectedIndexTensor,
+                                                      Rpp32s *detectionLengthTensor,
+                                                      Rpp32f cutOffDB,
+                                                      Rpp32s windowLength,
+                                                      Rpp32f referencePower,
+                                                      Rpp32s resetInterval,
+                                                      rpp::Handle& handle)
+{
+    // check if scratch memory size required for moving mean square is within the limits
+    if ((srcDescPtr->n * srcDescPtr->strides.nStride) > 76800000) // 76800000 is the maximum scratch memory size needed for MMS buffer in RNNT training
```
AryanSalmanpour, line 323: please #define this 76800000 number and use it here.

AryanSalmanpour (changes requested):
Please also add a CTest for this kernel.

include/rppdefs.h
```diff
@@ -64,6 +64,7 @@ SOFTWARE.
 const float ONE_OVER_6 = 1.0f / 6;
 const float ONE_OVER_3 = 1.0f / 3;
 const float ONE_OVER_255 = 1.0f / 255;
+const uint MMS_MAX_SCRATCH_MEMORY = 76800000; // maximum scratch memory size needed for MMS buffer in RNNT training
```
rrawther, line 67: Is this in bytes. Please specify and modify the name

src/modules/hip/hip_tensor_audio_augmentations.hpp
```diff
@@ -0,0 +1,30 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
+
+#ifndef HIP_TENSOR_AUDIO_AUGMENTATIONS_HPP
+#define HIP_TENSOR_AUDIO_AUGMENTATIONS_HPP
+
+#include "kernel/non_silent_region_detection.hpp"
```
rrawther, line 28: if this is only including one header file, you don't really need a separate header file. You can directly include kernel/non_silent_region_detection.hpp in where it is referrenced

rrawther (changes requested):

src/modules/hip/hip_tensor_audio_augmentations.hpp
sampath1117: Hi Rajy,
This header file is used for including all audio HIP kernel
Currently since this is the first audio HIP PR, this has only included only 1 HIP kernel
Once few more audio HIP kernels are merged, this header will have more included like it is for HOST
https://github.com/ROCm/rpp/blob/develop/src/modules/cpu/host_tensor_audio_augmentations.hpp

src/modules/hip/kernel/non_silent_region_detection.hpp
r-abishek: Changed to constant and CTest added here -> https://github.com/ROCm/rpp/pull/395/files#diff-df94f1927a53d83c734f3a0e7d81a532662ca9505bf1c1cc788597ed108489cdR125

include/rppdefs.h
r-abishek: Comment modified to clarify that it's the number of elements in the moving mean square array

AryanSalmanpour (approved):

rrawther (approved):

PR #399, opened 2024-07-12
RPP Test Suite - Replace fstrings and subprocess calls to support old/new versions of Python

r-abishek:
This PR adds relevant changes to support RPP test suite on old/new versions of python - tested on python 2.7, 3.5, 3.10
- Replaces usage of fstrings with just strings for python backward compatibility
- Replaces usage of subprocess.run with just subprocess.call / subprocess.Popen for python backward compatibility
- Minor bugfix in test suite help display


r-abishek:
@LakshmiKumar23 @kiritigowda This PR now passes CI and should support older and newer versions of Python.

LakshmiKumar23:
> @LakshmiKumar23 @kiritigowda This PR now passes CI and should support older and newer versions of Python.

I will update the ticket so QA can test and confirm

rrawther (approved):
@LakshmiKumar23 to run tests to see if the issue is fixed

PR #402, opened 2024-07-17
RPP Audio Support HIP - Pre emphasis filter

r-abishek:
- Adds RPP audio support for Pre emphasis filter on HIP backend.
- Adds relevant qa tests and performance tests.


kiritigowda:
@r-abishek need conflicts resolved

src/modules/hip/kernel/pre_emphasis_filter.hpp
```diff
@@ -0,0 +1,72 @@
...
+        return;
+
+    uint srcIdx = (id_z * srcStride) + id_x;
+    uint dstIdx = (id_z * dstStride) + id_x;
+    float coeff = coeffTensor[id_z];
+
+    d_float8 src1_f8, src2_f8, dst_f8;
+    float4 coeff_f4 = static_cast<float4>(coeff);
+    rpp_hip_load8_and_unpack_to_float8(srcPtr + srcIdx, &src1_f8);
+    if(id_x == 0)
```
rrawther, line 31: can't we merge the if id_x==0 case in advance before calling the kernel. It will block all threads

rrawther (changes requested):

src/modules/hip/kernel/pre_emphasis_filter.hpp
HazarathKumarM: Hi Rajy,

We  moved the if (id_x == 0) case outside the kernel and measured the performance. The change resulted in slower performance compared to the current approach. The condition only applies to the first eight threads per block, so it doesn't cause much delay. Also, the else part is independent of the if condition computation, so it won't block all the threads.

rrawther (approved):
ci has build issues I guess

kiritigowda:
> ci has build issues I guess

CI issue fixed

PR #405, opened 2024-07-24
RPP Audio Support HIP - Resample

r-abishek:
Adds RPP audio support for Resample on HIP backend.
Adds relevant qa tests and performance tests.


LakshmiKumar23 (approved):

src/modules/hip/kernel/resample.hpp
```diff
@@ -0,0 +1,298 @@
+#include <hip/hip_runtime.h>
+#include "rpp_hip_common.hpp"
+
+// -------------------- Set 0 - resample kernel device helpers  --------------------
+
+__device__ __forceinline__ float resample_hip_compute(float &x, float &scale, float &center, float *lookup, int &lookupSize)
+{
+    float locRaw = x * scale + center;
+    int locFloor = std::floor(locRaw);
+    float weight = locRaw - locFloor;
```
rrawther, line 10: suggest to combine these so compiler can optimize better
float weight = x * scale + center - std::floor(locRaw);

src/modules/hip/kernel/resample.hpp
```diff
@@ -0,0 +1,298 @@
...
+    locFloor = std::max(std::min(locFloor, lookupSize - 2), 0);
+    float current = lookup[locFloor];
+    float next = lookup[locFloor + 1];
```
rrawther, line 13: please combine line 12 and 13 and compute diff
float diff = lookup[locFloor + 1] - lookup[locFloor];

src/modules/hip/kernel/resample.hpp
```diff
@@ -0,0 +1,298 @@
...
+    return current + weight * (next - current);
+}
+
+__device__ __forceinline__ void resample_hip_compute(float4 *src_f4, float4 *dst_f4, const float4 *scale_f4, const float4 *center_f4, float *lookup)
```
rrawther, line 17: pass the argument by ref and use it directly inside the function

src/modules/hip/kernel/resample.hpp
```diff
@@ -0,0 +1,298 @@
...
+{
+    float4 locRaw_f4 = (*src_f4) * (*scale_f4) + (*center_f4);
+    int4 locFloor_i4 = make_int4(std::floor(locRaw_f4.x), std::floor(locRaw_f4.y), std::floor(locRaw_f4.z), std::floor(locRaw_f4.w));
```
rrawther, line 20: can't you use floor(locRaw_f4) instead?

utilities/test_suite/HOST/Tensor_audio_host.cpp
```diff
@@ -464,5 +469,7 @@ int main(int argc, char **argv)
     free(dstDims);
     free(inputf32);
     free(outputf32);
+    if (testCase == 6)
+        free(window.lookup);
```
rrawther, line 473: instead of doing that, it is better to check the pointer validity

rrawther (changes requested):

kiritigowda:
@r-abishek - need conflicts resolved

src/modules/hip/kernel/resample.hpp
sampath1117: Hi Rajy,
we are storing  x * scale + center in locRaw
and  locFloor = floor(locRaw)

This locFloor is again used in the next steps, which is why we stored x * scale + center and std::floor(locRaw) in separate variables
```
    locFloor = std::max(std::min(locFloor, lookupSize - 2), 0);
    float current = lookup[locFloor];
    float next = lookup[locFloor + 1];
```

src/modules/hip/kernel/resample.hpp
sampath1117: Hi Rajy,
The final return value from this function is 
```
current + weight * (next - current);
```

So if we store difference in separate variable as below
```
float diff = lookup[locFloor + 1] - lookup[locFloor];
```

Then we need to return final value as
return lookup[locFloor] + weight * diff

This requires the 3 memory access from lookup
to reduce this to 2 memory access, we stored them in separate variables in current code and return the final value as
```
    float current = lookup[locFloor];
    float next = lookup[locFloor + 1];
    return current + weight * (next - current);
```

src/modules/hip/kernel/resample.hpp
sampath1117: Hi Rajy,
locRaw_f4 is float4 variable 
Currently there is no function available in HIP that can do floor operation on float4 variables directly

rrawther: I suggest having a macro for floor(float4) and use hip math instructions instead of std::floor

rrawther:
Please remove merge conflict and address review comments

utilities/test_suite/HOST/Tensor_audio_host.cpp
sampath1117: Done

src/modules/hip/kernel/resample.hpp
r-abishek: macro for hip created 👍🏼

sampath1117:
@rrawther 
All the review comments has been addressed. Kindly note

include/rppdefs.h
```diff
@@ -61,6 +63,13 @@ SOFTWARE.
   } \
 } while (0)
 
+#ifdef HIP_COMPILE
```
rrawther, line 66: We need to get rid of this HIP_COMPILE flag and make it always enable

include/rppdefs.h
```diff
@@ -734,10 +743,48 @@ typedef struct RpptResamplingWindow
...
     Rpp32s lookupSize = 0;
-    std::vector<Rpp32f> lookup;
+    Rpp32f *lookup = nullptr;
     __m128 pCenter, pScale;
 } RpptResamplingWindow;
 
+inline Rpp32f sinc(Rpp32f x)
+{
+    x *= M_PI;
+    return (std::abs(x) < 1e-5f) ? (1.0f - x * x * (1.0f / 6)) : std::sin(x) / x;
```
rrawther, line 753: please change (1.f - (x * x * 0.16666667))

src/include/hip/rpp_hip_common.hpp
```diff
@@ -156,6 +156,14 @@ struct RPPTensorFunctionMetaData
   (byte & 0x02 ? '1' : '0'), \
   (byte & 0x01 ? '1' : '0')
 
+// float4 floor
+
+#define FLOOR4(src, dst) \
+dst.x = static_cast<int>(floorf(src.x)); \
```
rrawther, line 162: dst = make_float4(floorf(src.x), floorf(src.y), floorf(src.z), floorf(src.w)) would be enough here

rrawther (approved):
Added some comments

kiritigowda:
@r-abishek can you resolve the conflicts and address the comments.

include/rppdefs.h
r-abishek: Done

src/include/hip/rpp_hip_common.hpp

r-abishek:
@kiritigowda Ready to merge after CI

PR #410, opened 2024-07-31
RPP Pixelate on HOST and HIP

r-abishek:
- Adds support for Pixelate on HOST and HIP with bit-depth support for u8/f16/f32/i8.
- Adds relevant unit/performance/qa tests.


src/modules/rppt_tensor_effects_augmentations.cpp
```diff
@@ -1004,6 +1004,143 @@ RppStatus rppt_jitter_host(RppPtr_t srcPtr,
...
+                                          roiTensorPtrSrc,
+                                          roiType,
+                                          srcLayoutParams,
+                                          rpp::deref(rppHandle));
+        for(int i = 0; i < dstDescPtr->n; i++)
+        {
+            internalDstImgSizes[i].width = roiTensorPtrSrc[i].xywhROI.roiWidth;
+            internalDstImgSizes[i].height = roiTensorPtrSrc[i].xywhROI.roiHeight;
+        }
+        resize_nn_u8_u8_host_tensor(static_cast<Rpp8u*>(interDstPtr),
```
rrawther, line 1052: not sure why we are doing bilinear followed by nn

rrawther (approved):

PR #413, opened 2024-08-02
Improve finding half.hpp

trixirt:
On Fedora half.hpp is installed to the system dir /usr/include. There is no cmake package, so finding the half cmake package will always fail. The source looks for half with #include <half/half.h>  The use of half/ makes it impossible to find and use the system location.

So look for the half.hpp path and add add that to the include_directories.

Use this sed script to programatically change the uses of #include <half/half.h>

  for f in `find . -type f -name '*.hpp' -o -name '*.cpp' `; do
      sed -i -e 's@#include <half/half.hpp>@#include <half.hpp>@' $f
  done

This covers this issue https://github.com/ROCm/rpp/issues/292


rrawther (commented):
I think we need to include "half.hpp" in one common each in modules and utilities and inlude the common header file everywhere. If we did that before, we would have only 1/2 file changes in this PR :)

rrawther (approved):
I have added a comment for improving it in future by consolidating into a common file. Please incorporate in future

CMakeLists.txt
```diff
@@ -137,8 +137,9 @@ endif()
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 
 # Half
-find_package(HALF REQUIRED)
```
kiritigowda, line 140: @trixirt -- can you add this change in https://github.com/ROCm/rpp/blob/develop/cmake/FindHALF.cmake

kiritigowda:
@trixirt -- changes failing all CI tests

```
CMake Error in src/modules/CMakeLists.txt:
  Found relative path while evaluating include directories of "modules":

    "HALF_INCLUDE_DIR-NOTFOUND"



CMake Error in addkernels/CMakeLists.txt:
  Found relative path while evaluating include directories of "addkernels":

    "HALF_INCLUDE_DIR-NOTFOUND"



-- Generating done (0.0s)
CMake Warning:
  Manually-specified variables were not used by the project:

    HALF_INCLUDE_DIRS


CMake Generate step failed.  Build files cannot be regenerated correctly.

##[error]The process '/usr/local/bin/cmake' failed with exit code 1
##[error]CMake failed with error: The process '/usr/local/bin/cmake' failed with exit code 1

```

kiritigowda:
PR #459 - adds this fix

PR #414, opened 2024-08-02
Improve compiler setup

trixirt:
If the user has passed in the compiler to use for building, use it. Otherwise set it to clang++

On Fedora there is no /opt/rocm/llvm/bin/clang++, the compiler used is a system compiler wrapped by hipcc. So set COMPILER_FOR_HIP to llvm/bin/clang++ only if that file exists, otherwise use the CMAKE_CXX_COMPILER.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,9 @@ SOFTWARE.
 ]]
 
 cmake_minimum_required(VERSION 3.5)
-set(CMAKE_CXX_COMPILER clang++)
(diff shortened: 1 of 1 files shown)


rrawther (approved):

kiritigowda:
Merege pending - CI -- http://math-ci.amd.com/blue/organizations/jenkins/main%2Fprecheckin%2Frpp/detail/PR-414/1/pipeline

PR #428, opened 2024-08-13
Test Suite - updates

kiritigowda:
Updates

### DIFF
--- a/utilities/test_suite/README.md
+++ b/utilities/test_suite/README.md
@@ -1,49 +1,18 @@
-# AMD ROCm Performance Primitives (RPP) Test Suite
+# RPP Test Suite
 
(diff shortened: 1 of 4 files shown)


utilities/test_suite/README.md
```diff
@@ -59,37 +28,45 @@ This repository contains four test suites for the AMD ROCm Performance Primitive
   make -j$nproc
```
LakshmiKumar23, line 28: update the cmake flags to reflect this: https://github.com/ROCm/rocAL/blob/2f9a7fd4ef3b89acaf639592bce494907eb7b24a/rocAL-setup.py#L362
the current one doesn't work with the test suite

PR #438, opened 2024-08-26
Docs fix - Match parameter name in API and Docs

r-abishek:
- Fix roiTensorSrc to roiTensorPtrSrc in the header files to ensure parameter name is same as documentation name and avoid Doxygen warnings


randyh62 (approved):
Looks good to me. Does roiTensorScr actually exist in any context?

PR #442, opened 2024-08-27
MIVisionX compatibility fix - Resample and preemphasis filter

r-abishek:
- Move out the windowed_sinc function from rppdefs.h and readded vector usage for lookup table for compatibility
- Disable the numDims validation for preemphasis HOST kernel temporarily

### DIFF
--- a/utilities/test_suite/rpp_test_suite_audio.h
+++ b/utilities/test_suite/rpp_test_suite_audio.h
@@ -317,3 +317,45 @@ void verify_non_silent_region_detection(int *detectedIndex, int *detectionLength
         qaResults.close();
     }
 }
+
+inline Rpp32f sinc(Rpp32f x)
+{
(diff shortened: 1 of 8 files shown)


utilities/test_suite/rpp_test_suite_audio.h
```diff
@@ -316,4 +316,46 @@ void verify_non_silent_region_detection(int *detectedIndex, int *detectionLength
         qaResults << status << std::endl;
         qaResults.close();
     }
+}
+
+inline Rpp32f sinc(Rpp32f x)
```
rrawther, line 321: when we enable RPP HIP audio, we need to remove these from here and MIVisionX. To do that you can use a new RPP version and check that

rrawther (approved):
Added some comment for future improvement

PR #448, opened 2024-09-13
Bugfix - RPP normalize generic

r-abishek:
Adds bugfix to match generic kernel output to golden output for normalize


zacharyvincze (approved):
Bug fix confirmed working on minimal example.

PR #463, opened 2024-10-23
RPP Rain augmentation on HOST and HIP

r-abishek:
- Adds to version to RPP 1.9.10
- Adds a dash-cam rain augmentation on HOST and HIP
- Adds support for U8/F16/F32/I8 bit depths with NCHW/NHWC tensor support


utilities/test_suite/HIP/runTests.py
```diff
@@ -478,7 +478,7 @@ def rpp_test_suite_parser_and_validator():
         print_performance_tests_summary(logFile, functionalityGroupList, numRuns)
 
 # print the results of qa tests
-nonQACaseList = ['6', '8', '24', '54', '84'] # Add cases present in supportedCaseList, but without QA support
```
rrawther, line 481: does this also has precision difference? why is this not added for QA?

utilities/test_suite/HOST/Tensor_host.cpp
```diff
@@ -67,7 +67,7 @@ int main(int argc, char **argv)
     bool additionalParamCase = (testCase == 8 || testCase == 21 || testCase == 23 || testCase == 24 || testCase == 49 || testCase == 79);
     bool kernelSizeCase = (testCase == 49);
     bool dualInputCase = (testCase == 2 || testCase == 30 || testCase == 33 || testCase == 61 || testCase == 63 || testCase == 65 || testCase == 68);
-    bool randomOutputCase = (testCase == 6 || testCase == 8 || testCase == 84);
```
rrawther, line 70: I have commented many times before. We should not use numbers for test cases. It is better to use ENUMs here

utilities/test_suite/HOST/runTests.py
```diff
@@ -260,7 +260,7 @@ def rpp_test_suite_parser_and_validator():
 subprocess.call(["make", "-j16"], cwd=".")    # nosec
 
 # List of cases supported
-supportedCaseList = ['0', '1', '2', '4', '5', '6', '8', '13', '20', '21', '23', '26', '29', '30', '31', '32', '33', '34', '35', '36', '37', '38', '39', '45', '46', '49', '54', '61', '63', '65', '68', '70', '79', '80', '81', '82', '83', '84', '85', '86', '87', '88', '89', '90', '91', '92']
```
rrawther, line 263: same as before. Need to change numbers to ENUMs

utilities/test_suite/rpp_test_suite_common.h
```diff
@@ -154,4 +154,4 @@ inline void read_bin_file(string refFile, T *binaryContent)
     fseek(fp, 0, SEEK_SET);
     fread(binaryContent, fsize, 1, fp);
     fclose(fp);
-}
+}
```
rrawther, line 157: please revert this file as there seems to be no change

CHANGELOG.md
```diff
@@ -2,9 +2,16 @@
 
 Full documentation for RPP is available at [https://rocm.docs.amd.com/projects/rpp/en/latest](https://rocm.docs.amd.com/projects/rpp/en/latest)
 
+## (Unreleased) RPP 1.9.8
+
+### Changes
```
spolifroni-amd, line 7:
```suggestion
### Changed
```

CHANGELOG.md
```diff
@@ -2,9 +2,16 @@
...
+
+RPP Rain augmentation on HOST and HIP
+
 ## (Unreleased) RPP 1.9.4
 
 ### Changes
+
```
spolifroni-amd, line 14: Is this also unreleased? Because 1.9.8 is unreleased.

spolifroni-amd (changes requested):

kiritigowda:
@r-abishek - needs conflicts resolved

utilities/test_suite/HIP/Tensor_hip.cpp
```diff
@@ -432,6 +432,10 @@ int main(int argc, char **argv)
     if(testCase == 28)
```
rrawther, line 432: we should use ENUM for test case instead of hardcoded values for better code readability

utilities/test_suite/HOST/Tensor_host.cpp
r-abishek: Addressed with #499

utilities/test_suite/HOST/runTests.py
r-abishek: Addressed with #499

utilities/test_suite/HIP/Tensor_hip.cpp
r-abishek: Addressed with #499

utilities/test_suite/HIP/runTests.py
r-abishek: These are functions that involve random rain/fog/snow/noise that don't have the pixel-wise match test.

rrawther (approved):

spolifroni-amd (approved):
The points in the changed section need to be changed so they have more info (what was it before; what is it now; what, if anything, does the user need to do)

But I don't want to hold up the review process so I'll OK this as is and we can change it later.

kiritigowda:
@r-abishek can you look at the conflicts

r-abishek:
@kiritigowda Could we merge this PR? 
Conflicts resolved, Azure passing.

PR #465, opened 2024-10-30
Bugfix - RPP BitwiseAND BitwiseOR

r-abishek:
- Fix mismatches in output between FP16 and other bit depths due to precision differences.
- Use rounding off operation instead of truncation.

### DIFF
--- a/src/include/hip/rpp_hip_common.hpp
+++ b/src/include/hip/rpp_hip_common.hpp
@@ -1802,28 +1802,28 @@ __device__ __forceinline__ void rpp_hip_math_divide8_const(d_float8 *src_f8, d_f
 
 __device__ __forceinline__ void rpp_hip_math_bitwiseAnd8(d_float8 *src1_f8, d_float8 *src2_f8, d_float8 *dst_f8)
 {
-        dst_f8->f1[0] = (float)((uchar)(src1_f8->f1[0]) & (uchar)(src2_f8->f1[0]));
-        dst_f8->f1[1] = (float)((uchar)(src1_f8->f1[1]) & (uchar)(src2_f8->f1[1]));
-        dst_f8->f1[2] = (float)((uchar)(src1_f8->f1[2]) & (uchar)(src2_f8->f1[2]));
-        dst_f8->f1[3] = (float)((uchar)(src1_f8->f1[3]) & (uchar)(src2_f8->f1[3]));
-        dst_f8->f1[4] = (float)((uchar)(src1_f8->f1[4]) & (uchar)(src2_f8->f1[4]));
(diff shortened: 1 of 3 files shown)


src/include/hip/rpp_hip_common.hpp
```diff
@@ -1802,28 +1802,28 @@ __device__ __forceinline__ void rpp_hip_math_divide8_const(d_float8 *src_f8, d_f
...
 {
-        dst_f8->f1[0] = (float)((uchar)(src1_f8->f1[0]) & (uchar)(src2_f8->f1[0]));
-        dst_f8->f1[1] = (float)((uchar)(src1_f8->f1[1]) & (uchar)(src2_f8->f1[1]));
-        dst_f8->f1[2] = (float)((uchar)(src1_f8->f1[2]) & (uchar)(src2_f8->f1[2]));
-        dst_f8->f1[3] = (float)((uchar)(src1_f8->f1[3]) & (uchar)(src2_f8->f1[3]));
-        dst_f8->f1[4] = (float)((uchar)(src1_f8->f1[4]) & (uchar)(src2_f8->f1[4]));
-        dst_f8->f1[5] = (float)((uchar)(src1_f8->f1[5]) & (uchar)(src2_f8->f1[5]));
-        dst_f8->f1[6] = (float)((uchar)(src1_f8->f1[6]) & (uchar)(src2_f8->f1[6]));
-        dst_f8->f1[7] = (float)((uchar)(src1_f8->f1[7]) & (uchar)(src2_f8->f1[7]));
+        dst_f8->f1[0] = (float)((uchar)(std::nearbyintf)(src1_f8->f1[0]) & (uchar)(std::nearbyintf)(src2_f8->f1[0]));
```
rrawther, line 1805: This is wrong as it will change the float value before doing AND. Please do reinterpret_cast<int> instead

rrawther (changes requested):

src/include/hip/rpp_hip_common.hpp
Srihari-mcw: - The logical operations (XOR, Bitwise And etc) support is provided for images only. Images can be stored in the following formats and carry the following values for the formats : U8 (0 to 255), I8 (-128 to 127), Float (0 to 1)
- In case of float image, we rescale the images to 0 to 255 range before we perform the logical operations. While rescaling, some of the values are close to the ceil value (Eg - 224.99267578125 in case of representing 225) and to account for the same rounding operation is done here. Usage of reinterpret cast and performing the operation leads to output mismatches in this case. Thanks

rrawther: What is the purpose of these operators? Is this supposed to do logical or arithmetical operation?

rrawther (changes requested):
Added a comment

r-abishek:
All changes to Bitwise AND and OR as part of #520 
Closing.

PR #466, opened 2024-10-30
updated the changelog for 6.3

spolifroni-amd:


rrawther:
@r-abishek : can you please review this

CHANGELOG.md
```diff
@@ -2,74 +2,63 @@
...
+## (Unreleased) RPP 1.9.4
 ### Changes
 
 * RPP Tensor Box Filter support on HOST
 
-## RPP 1.9.3 (unreleased)
+## (Unreleased) RPP 1.9.3
 
-### Changes
+### Changed
```
r-abishek, line 12: @spolifroni-amd If we adopt the "Changes"->"Changed" fix, we need to change that for all other instances of "Changes" in this file?

Also, if we want to change "## RPP 1.9.3 (unreleased)" to "## (Unreleased) RPP 1.9.3", do we want to adopt that format for all other other open PRs like 446, 451, 456, 463 and 464?

spolifroni-amd: The "changes" -> "changed" only applies to 6.3 and onwards. Don't worry about past versions.

re "RPP (unreleased)" -> "(Unreleased) RPP", that change needs to be made if the PRs impact 6.3. Leaving it as-is will interfere with the release notes script used by the documentation team. If the PRs don't impact 6.3 -- even if they impact later versions -- then it can be left as-is for now and we'll revisit at the next release. But this needs to be fixed in the release-6.3 branch.

r-abishek (approved):

CHANGELOG.md
LakshmiKumar23: @spolifroni-amd @r-abishek why is it unreleased? The changes for rocm6.3 would be released right?

spolifroni-amd: Unreleased is used for upcoming versions. The topmost item seems to be for an upcoming version, so it would be labelled Unreleased and not be picked up by the release notes script.

the 6.3 item doesn't have (Unreleased) in front of it, so it will be picked up by the script.

amd-jnovotny (approved):

PR #479, opened 2024-11-20
Bugfix - Log Tensor

r-abishek:
Bugfix for log tensor in stride updation in log_recursive function


src/modules/cpu/kernel/log.hpp
```diff
@@ -44,8 +44,8 @@ void log_recursive(T1 *src, Rpp32u *srcStrides, T2 *dst, Rpp32u *dstStrides, Rpp
         for (int i = 0; i < *dstShape; i++)
         {
             log_recursive(src, srcStrides + 1, dst, dstStrides + 1, dstShape + 1, nDim - 1);
-            dst += *dstStrides;
-            src += *srcStrides;
+            dst += *(dstStrides + 1);
```
rrawther, line 47: I don't understand the logic here

rrawther (changes requested):
please address the review comment

src/modules/cpu/kernel/log.hpp
Srihari-mcw: - The fix attempts to update the stride for respective dimensions. The log_recursive function recursively calls itself until the lowest dimension (of shape 1) where the destination value is set to the log of source value.
- For eg, if the strides[1] = 16384 and strides[2] = 128, instead of updating with strides[2] i.e 128, the pointer gets updated by 16384 which leads to out of memory access
- The earlier code led to the stride getting updated for higher dimension which led to memory access issues. The current code however, post the log_recursive call updates the pointers with stride for respective dimension. Thanks

Srihari-mcw: - The fix attempts to update the stride for respective dimensions. The log_recursive function recursively calls itself until the lowest dimension (of shape 1) where the destination value is set to the log of source value.
- The earlier code led to the stride getting updated for higher dimension which led to memory access issues.
- For eg, if the strides[1] = 16384 and strides[2] = 128, instead of updating with strides[2] i.e 128, the pointer gets updated by 16384 which leads to out of memory access
-  The current code however, post the log_recursive call updates the pointers with stride for respective dimension. Thanks

rrawther (approved):

kiritigowda (approved):

PR #480, opened 2024-11-23
Compiler - updates and cleanup

kiritigowda:
* Use AMD Clang++ 
* CMake Version 3.10


LakshmiKumar23 (approved):
Can merge if all tests pass

randyh62 (approved):

r-abishek (approved):
Looks good, Host shouldn't be affected. 
Assuming hip kernel compilation is also not affected in changing hipcc to amdclang++.

Should be good if all CI HIP kernel runs pass.

spolifroni-amd (approved):

PR #507, opened 2025-01-16
Test Package - Updates

kiritigowda:
Test package to bring most of it's dependencies

### DIFF
--- a/utilities/test_suite/README.md
+++ b/utilities/test_suite/README.md
@@ -4,16 +4,6 @@ This repository contains four test suites for the RPP library: `image`/`voxel`/`
 
 ## Prerequisites
 
-* OpenMP
-  ```shell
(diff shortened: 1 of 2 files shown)


spolifroni-amd (approved):

utilities/test_suite/README.md
```diff
@@ -4,16 +4,6 @@ This repository contains four test suites for the RPP library: `image`/`voxel`/`
 
 ## Prerequisites
 
-* OpenMP
-  ```shell
-  sudo apt install libomp-dev
```
r-abishek, line 9: We are removing explicitly stating OpenMP in the deps?

kiritigowda: Yes, we will use AOMP - AMD provided OpenMP which will get installed with ROCm

PR #520, opened 2025-02-19
RPP Tensor Support - Bitwise NOT, Bitwise AND, Bitwise OR

r-abishek:
- Add support for Bitwise NOT for U8.
- Restrict bitwise ops (AND OR NOT) to U8.
- Rename ops from 'logical' to 'bitwise'.


spolifroni-amd (commented):
Does this need to be addressed in the changelog? If not, it's fine from a docs point-of-view.

rrawther (approved):

spolifroni-amd (approved):

PR #525, opened 2025-03-05
RPP Major Restructure and Compile Time Reduction

r-abishek:
The PR contains a major code-structure update to RPP without changing any functionality.
The PR changes have no effect whatsoever from a user's standpoint / RPP API usage standpoint.

Changes include:
1. Significant compile time reduction:
    - A complete re-design of RPP folder structure to achieve a **~44% faster compile time on HIP backend**, and about **~48% faster compile time on HOST backend**.
2. Build parallelization efficiently utilizing CPU threads:
    - Changes to RPP Tensor codebase to build **separate object files for every HIP kernel and every HOST kernel** so that 'make' parallelizes the build.
3. Major code-structure reorganization:
    - The outer `include` folder containing all external RPP API to now be called `api`.
    - Complete isolation of folders containing the BatchPD codebase vs the Tensor codebase inside `rpp/src/include` as well as `rpp/src/modules`.
    - Structuring all common code used by BatchPD and Tensor codebases under `rpp/src/include/common`.
    - Ensuring every functionality has its own `<kernel_name>.cpp` file for both HIP and HOST backends, and only includes the minimum necessary files for build.
    - Moving any helper functions specific to only a single kernel into the `<kernel_name>.cpp` file.
    - Splitting the long headers `rpp_cpu_common.hpp`, `rpp_cpu_simd.hpp` on HOST backend, and `rpp_hip_common.hpp` on HIP backend into smaller header files grouped logically by category.
    - Remove unnecessary or redundant folders and files.

<p align="center">image</p>

<p align="center">image</p>


LakshmiKumar23:
@r-abishek please coordinate to make the corresponding change in other places where `rpp/include/` is used. For example, the `rpp_backend.h` [file](https://github.com/ROCm/rpp/pull/525/files#diff-807a0dd972e622f616e6ebb11dc5d46724a59cb79d0d1dad583aa807d2d56886) is used in rocAL tests [here](https://github.com/ROCm/rocAL/blob/c3d72634ed1b86eac0b6dc83201bb4b9f3948624/tests/cpp_api/CMakeLists.txt#L64).

r-abishek:
@LakshmiKumar23 We'll reconfirm the builds post checking MIVisionX and rocAL and their tests.

src/modules/batch_pd/cl/hip_build_utils.cpp
LakshmiKumar23, line 1: Why is a hip file being moved into cl/ directory?

r-abishek: Older code with same issue in ToT, so it hasn't been touched in this PR - https://github.com/ROCm/rpp/blob/develop/src/modules/cl/hip_build_utils.cpp

Should be a separate PR to evaluate/fix/remove BatchPD code.

r-abishek:
@rrawther @LakshmiKumar23 @kiritigowda 
All changes re-tested with new updates - Passing on CI and local:
1. Moved third_party outside api folder. External folder is now called rpp/third_party. It has include and lib.
2. Removed export.h file and build.
3. Removed filesystem.h and added to rpp.h.
4. rppt_api_helpers now moved inside instead of in external api folder.
5. Significant reduction in the include hpp files by combining all src/include/cpu/kernel/* into one file for host and one for hip. See a 4-8% compilation time reduction on it but overall gains still above 40% and there's lesser files.

LakshmiKumar23 (approved):

rrawther (approved):

PR #528, opened 2025-03-22
MIVisionX compatibility fix - Filesystem import updates

Srihari-mcw:
The objective of the PR is to update the filesystem import in rpp.
Due to the recent changes in restructure PR related to Commit - [Remove filesystem.h file and consolidate in rpp.h](https://github.com/r-abishek/rpp/commit/bc2c36ebc5859ba027dca2c664bff9e449c0048c) the MIVisionX build started to fail

This issue is owing to the fact that MIVisionX for rpp dependencies uses CXX Standard 14 contrary to RPP that uses CXX 17 for compilation - https://github.com/ROCm/MIVisionX/blob/c166676dfe5e9bf1c356554a244bde4d7f4ebb8f/amd_openvx_extensions/amd_rpp/CMakeLists.txt#L25

Hence changes have been added to restore support for multiple CXX standards in this commit

### DIFF
--- a/api/rpp.h
+++ b/api/rpp.h
@@ -51,11 +51,11 @@ typedef hipStream_t rppAcceleratorQueue_t;
 #endif
 
 #if __cplusplus >= 201703L && __has_include(<filesystem>)
-    #include <filesystem>
-    namespace fs = std::filesystem;
+#include <filesystem>
+namespace fs = std::filesystem;
 #elif __has_include(<experimental/filesystem>)
-    #include <experimental/filesystem>
-    namespace fs = std::experimental::filesystem;
+#include <experimental/filesystem>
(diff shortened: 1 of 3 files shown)


api/rpp.h
```diff
@@ -50,8 +50,13 @@ typedef hipStream_t rppAcceleratorQueue_t;
 
 #endif
 
+#if __cplusplus >= 201703L && __has_include(<filesystem>)
```
rrawther, line 53: @r-abishek : Do we need to include filesystem in the main header. We shouldn't include here. Can we move this to an internal file

kiritigowda (approved):

rrawther (approved):

PR #536, opened 2025-04-03
changed the way images are coded

spolifroni-amd:


rrawther (commented):
Why is this changed?

spolifroni-amd:
> Why is this changed?

When the rst was being converted to html, the images were being interpreted as links for some reason. Two still are despite the change.  I'll have to change the size of those images so that they don't cause problems with the rst->html conversion.

rrawther (approved):

kiritigowda (approved):

PR #549, opened 2025-04-24
Test suite - Add QA pass/fail tests for F32 bit depth

r-abishek:
The PR adds internal QA testing support for F32 bit depth for next set of functions.
- Add range checks rpp_pixel_check_0to1 to ensure F32 bit depth QA test pass.
- Add necessary bin files for reference output.


LakshmiKumar23 (approved):

src/modules/tensor/cpu/kernel/color_twist.cpp
```diff
@@ -655,11 +655,15 @@ RppStatus color_twist_f32_f32_host_tensor(Rpp32f *srcPtr,
                     __m256 p[3];
                     rpp_simd_load(rpp_load24_f32pkd3_to_f32pln3_avx, srcPtrTemp, p);    // simd loads
                     compute_color_twist_24_host(p[0], p[1], p[2], pColorTwistParams);    // color_twist adjustment
+                    //Boundary checks for f32
+                    rpp_pixel_check_0to1(p, 3);
```
rrawther, line 659: why is this added now?

src/include/common/cpu/rpp_cpu_simd_load_store.hpp
```diff
@@ -273,16 +273,16 @@ inline void rpp_mm256_print_ps(__m256 vPrintArray)
     }
```
rrawther, line 273: I suggest using a union to represent vPrintArray with __m256 and float[8], that way copying the _m256 value changes the array as well

rrawther (changes requested):
Added few comments

rrawther (changes requested):
As we have discussed, please add the pixel check only in the functions where it is needed.

HazarathKumarM:
image

rrawther (commented):
@r-abishek  there was a comment which is not resolved

src/include/common/cpu/rpp_cpu_simd_load_store.hpp
HazarathKumarM: used union and removed the store instructions

src/modules/tensor/cpu/kernel/color_twist.cpp
r-abishek: this func needed the pixel check

r-abishek:
@rrawther All comments resolved now

PR #550, opened 2025-04-25
F16 variants - Update loads and stores to AVX2

r-abishek:
- Replacement of scalar load/store and conversion to FP32, with AVX2 intrinsics - no additions or removals to external user API.
- 5.33%-21.05% improvements in performance for the updated kernels for the FP16 bit depth.
- F16 Load/Store updates for set of 8 kernels - color_temperature, color_to_greyscale, color_twist, non_linear_blend, noise_gaussian, noise_shot, noise_salt_and_pepper, phase.
- Fixes to SSE version of color_twist and color_to_greyscale.


rrawther (changes requested):
I am assuming there was a bug in the previous version of this PR. Please add in the changelist what is fixed in this PR. Also I would assume we need to test both cases with and without AVX

rrawther (approved):
Can be merged after making sure the tests are passing

Srihari-mcw:
Hi @rrawther 

> I am assuming there was a bug in the previous version of this PR. Please add in the changelist what is fixed in this PR. Also I would assume we need to test both cases with and without AVX

The earlier version had F16 inputs casted into F32 in a scalar loop during load, and F32 SIMD registers casted back into F16 outputs during store. The current version properly uses AVX2 to load F16 inputs that gives better performance.
 
Also, the raw C, SSE and AVX Versions were all tested

Thanks

LakshmiKumar23 (approved):

PR #552, opened 2025-05-20
Test Suite - Change python scripts to use augmentation names instead of case numbers

r-abishek:
- Changes the test suite python scripts to use only augmentation names instead of case numbers for readability.
- Adds 2 missing erode and dilate test cases.


rrawther (changes requested):
I think this needs to add in changelog.md and bump minor version

LakshmiKumar23 (approved):

r-abishek:
@rrawther This doesn't impact the end user so didn't add it in the changelog.
The changes in python scripts are just for better script-readability.

PR #570, opened 2025-06-18
F16 variants - Update loads and stores to AVX2

r-abishek:
- Replacement of scalar load/store and conversion to FP32, with AVX2 intrinsics - no additions or removals to external user API.
- 12.25%-24.78% improvements in performance for the updated kernels for the FP16 bit depth.
- F16 Load/Store updates for set of 5 kernels - tensor_max, tensor_min, tensor_sum, tensor_mean, tensor_stddev.

image


src/modules/tensor/cpu/kernel/tensor_stddev.cpp
```diff
@@ -859,7 +843,7 @@ RppStatus tensor_stddev_f16_f32_host(Rpp16f *srcPtr,
             tensorStddevArr[idx] = stddevR;
             tensorStddevArr[idx + 1] = stddevG;
             tensorStddevArr[idx + 2] = stddevB;
-            tensorStddevArr[idx + 3] = stddevImage;
+            tensorStddevArr[idx + 3] = stddevImage;
```
rrawther: remove extra white space in this line

rrawther (changes requested):
consider copilot suggestion for using helper function

src/modules/tensor/cpu/kernel/tensor_stddev.cpp
Srihari-mcw: Resolved

rrawther (approved):

PR #575, opened 2025-06-28
Fix half.hpp path

utk4r-sh:
This commit adds half.hpp path for Debian systems.
@cgmb @pfultz2 please review.


cgmb:
@utk4r-sh, please target the develop branch by default. If the change is believed to be important enough to merit being backported into a patch release, it will be cherry-picked by the rpp development team into a release branch.

PR #576, opened 2025-07-01
Fix half.hpp path

utk4r-sh:
This commit adds half.hpp path for Debian systems.
@cgmb @pfultz2 please review.


r-abishek (approved):
Looks good if build passing

Issue #590, opened 2025-07-31
[Issue]: Compile Warnings - Azure CI


kiritigowda:
### Problem Description

https://dev.azure.com/ROCm-CI/ROCm-CI/_build/results?buildId=41543&view=logs&j=727a0067-9911-5584-ebf1-ed58131a5840&t=8c2d5ee4-e455-5dd0-bb6c-f2a51b0297bc

```

/home/vsts/work/1/s/src/include/common/rppi_validate.hpp:155:5: warning: ignoring return value of type 'hipError_t' declared with 'nodiscard' attribute [-Wunused-value]
  155 |     hipMemcpy(handle.GetInitHandle()->mem.mgpu.roiPoints.y, handle.GetInitHandle()->mem.mgpu.croiPoints.y, sizeof(Rpp32u) * handle.GetBatchSize(), hipMemcpyHostToDevice);
      |     ^~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/vsts/work/1/s/src/include/common/rppi_validate.hpp:171:5: warning: ignoring return value of type 'hipError_t' declared with 'nodiscard' attribute [-Wunused-value]
  171 |     hipMemcpy(handle.GetInitHandle()->mem.mgpu.floatArr[paramIndex].floatmem, handle.GetInitHandle()->mem.mcpu.floatArr[paramIndex].floatmem, sizeof(Rpp32f) * handle.GetBatchSize(), hipMemcpyHostToDevice);
      |     ^~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #606, opened 2025-09-03
Warning fixes for HIP Backend - Check return status

r-abishek:
Fixes "Compile Warnings - Azure CI" #590 
- Contains warning fixes for the HIP Backend build
- Adds CHECK_RETURN_STATUS for the various hip API calls


src/modules/tensor/hip/kernel/crop_and_patch.cpp
```diff
@@ -334,8 +334,8 @@ RppStatus hip_exec_crop_and_patch_tensor(T *srcPtr1,
 
     if ((srcDescPtr->layout == RpptLayout::NHWC) && (dstDescPtr->layout == RpptLayout::NHWC))
     {
-        hipMemcpyAsync(dstPtr, srcPtr2, static_cast<size_t>(srcDescPtr->n * srcDescPtr->strides.nStride * sizeof(T)), hipMemcpyDeviceToDevice, handle.GetStream());
-        hipStreamSynchronize(handle.GetStream());
+        CHECK_RETURN_STATUS(hipMemcpyAsync(dstPtr, srcPtr2, static_cast<size_t>(srcDescPtr->n * srcDescPtr->strides.nStride * sizeof(T)), hipMemcpyDeviceToDevice, handle.GetStream()));
+        CHECK_RETURN_STATUS(hipStreamSynchronize(handle.GetStream()));
         hipLaunchKernelGGL(crop_and_patch_pkd_hip_tensor,
```
rrawther, line 339: hipLaunchKernelGGL also needs to check the status after the call

src/include/common/rppi_validate.hpp
```diff
@@ -104,8 +104,8 @@ inline void copy_srcSize(RppiSize *srcSize, rpp::Handle& handle)
            handle.GetInitHandle()->mem.mgpu.csrcSize.width[i] = srcSize[i].width;
     }
 #ifdef HIP_COMPILE
-    hipMemcpy(handle.GetInitHandle()->mem.mgpu.srcSize.height, handle.GetInitHandle()->mem.mgpu.csrcSize.height, sizeof(Rpp32u) * handle.GetBatchSize(), hipMemcpyHostToDevice);
-    hipMemcpy(handle.GetInitHandle()->mem.mgpu.srcSize.width, handle.GetInitHandle()->mem.mgpu.csrcSize.width, sizeof(Rpp32u) * handle.GetBatchSize(), hipMemcpyHostToDevice);
+    CHECK_RETURN_STATUS(hipMemcpy(handle.GetInitHandle()->mem.mgpu.srcSize.height, handle.GetInitHandle()->mem.mgpu.csrcSize.height, sizeof(Rpp32u) * handle.GetBatchSize(), hipMemcpyHostToDevice));
```
rrawther, line 107: this function is exiting with -1. If this is called from RPP funcion call, it should print an error and return status

src/modules/legacy_dep/hip/kernel/snow.cpp
```diff
@@ -153,7 +153,7 @@ RppStatus snow_hip_batch(Rpp8u *srcPtr, Rpp8u *dstPtr, rpp::Handle& handle, Rppi
         plnpkdind = 3;
     Rpp32u max_height, max_width;
     max_size(handle.GetInitHandle()->mem.mgpu.csrcSize.height, handle.GetInitHandle()->mem.mgpu.csrcSize.width, handle.GetBatchSize(), &max_height, &max_width);
-    hipMemcpy(dstPtr, srcPtr, nbatchSize * max_height * max_width * channel * sizeof(unsigned char), hipMemcpyDeviceToDevice);
+    CHECK_RETURN_STATUS(hipMemcpy(dstPtr, srcPtr, nbatchSize * max_height * max_width * channel * sizeof(unsigned char), hipMemcpyDeviceToDevice));
```
rrawther, line 156: we want to return an error code instead of exit(-1)

rrawther (changes requested):

src/modules/tensor/hip/kernel/crop_and_patch.cpp
r-abishek: The warnings in CI and #590 for some reason came only from all HIP API other than hipLaunchKernelGGL().

We have another PR in the pipeline since there are many calls to hipLaunchKernelGGL(), and this change won't be tiedd to the #590 issue.
This change will be a separate PR

src/include/common/rppi_validate.hpp
r-abishek: These edits in this file are only to avoid warnings, they are not used in the tensor code at all and will be deprecated completely.
So for now, planning to keep the rppi_validate.hpp and the below comment inside file legacy_dep/hip/kernel/snow.cpp too as is.
Passing the same return status all the way up back to the user could be cleaned up in a separate PR.

src/modules/legacy_dep/hip/kernel/snow.cpp
r-abishek: same comment as above

src/modules/tensor/hip/kernel/tensor_sum.cpp
```diff
@@ -1212,7 +1212,7 @@ RppStatus hip_exec_tensor_sum(T *srcPtr,
         Rpp32u partialSumArrLength = gridDim_x * gridDim_y * gridDim_z;
         float *partialSumArr;
         partialSumArr = handle.GetInitHandle()->mem.mgpu.scratchBufferHip.floatmem;
```
rrawther, line 1214: partialSumArr gets allocated in scratch_mem without even checking if there is enough memory available. This is very dangerous and can cause mem allocation failures. We need to get rid of this kind of memory allocation. We would like to catch failure on allocation and gracefully exit if it happens instead of not assuming any bounds for memory

rrawther (commented):
Added a comment regarding the use of scratch memory. Need to have a better mechanism

kiritigowda:
All warnings fixed on AzureCI
```
[167/175] Building CXX object src/modules/CMakeFiles/modules.dir/tensor/cpu/kernel/water.cpp.o
[168/175] Building CXX object src/modules/CMakeFiles/modules.dir/legacy_dep/hip/kernel/fisheye.cpp.o
[169/175] Building CXX object src/modules/CMakeFiles/modules.dir/legacy_dep/hip/kernel/hue_saturation.cpp.o
[170/175] Building CXX object src/modules/CMakeFiles/modules.dir/legacy_dep/hip/kernel/snow.cpp.o
[171/175] Building CXX object src/modules/CMakeFiles/modules.dir/legacy_dep/cpu/kernel/fisheye.cpp.o
[172/175] Building CXX object src/modules/CMakeFiles/modules.dir/legacy_dep/cpu/kernel/hue_saturation.cpp.o
[173/175] Building CXX object src/modules/CMakeFiles/modules.dir/legacy_dep/cpu/kernel/snow.cpp.o
[174/175] Linking CXX shared library lib/librpp.so.2.1.0
[175/175] Creating library symlink lib/librpp.so.2 lib/librpp.so
```

kiritigowda (approved):

rrawther (approved):

PR #654, opened 2025-12-22
RPP Tensor Fisheye on HOST and HIP

Srihari-mcw:
- Adds Fisheye augmentation on HIP and HOST
- Adds support for U8/F32/F16/I8 bit depths and NCHW/NHWC variants with toggle support
- Adds relevant unit / qa / performance tests


api/rppt_tensor_geometric_augmentations.h
```diff
@@ -722,6 +722,9 @@ RppStatus rppt_transpose_host(RppPtr_t srcPtr, RpptGenericDescPtr srcGenericDesc
 RppStatus rppt_transpose_gpu(RppPtr_t srcPtr, RpptGenericDescPtr srcGenericDescPtr, RppPtr_t dstPtr, RpptGenericDescPtr dstGenericDescPtr, Rpp32u *permTensor, Rpp32u *roiTensor, rppHandle_t rppHandle);
 #endif // GPU_SUPPORT
 
+/*! @}
+ */
```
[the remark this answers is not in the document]
HazarathKumarM: Done

src/modules/tensor/cpu/kernel/fisheye.cpp
```diff
@@ -0,0 +1,989 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
+
+#include "host_tensor_executors.hpp"
+#include "rpp_cpu_simd_math.hpp"
+#include "rpp_cpu_interpolation.hpp"
+
+#if __AVX2__
```
rrawther, line 29: I think gcc/clang compilers support built-in function for checking this. __builtin_cpu_supports("avx2"). THis is preferrable for distributing the code because it checks at runtime. We can make another PR to change this for all the RPP code

src/modules/tensor/hip/kernel/fisheye.cpp
```diff
@@ -0,0 +1,356 @@
...
+#include "rpp_hip_math.hpp"
+#include "rpp_hip_interpolation.hpp"
+
+// -------------------- Set 0 - fisheye kernel device helpers --------------------
+
+__device__ void fisheye_srcloc_hip_compute(int index, int2 *widthHeight_i2, int4 *srcRoiPtr_i4, d_float8 *normX_f8,
+                                           d_float8 *normY_f8, d_float8 *dist_f8, d_float16 *locSrc_f16)
+{
+    float dist = dist_f8->f1[index];
+    if ((dist >= 0.0) && (dist <= 1.0))
```
rrawther, line 36: what happens if it doesn't fit in this range?

src/modules/tensor/hip/kernel/fisheye.cpp
```diff
@@ -0,0 +1,356 @@
...
+    {
+        float distNew = sqrtf(1.0 - dist * dist);
+        distNew = (dist + (1.0 - distNew)) * 0.5f;
+        if (distNew <= 1.0)
```
rrawther, line 41: same comment here

rrawther (changes requested):

src/modules/tensor/hip/kernel/fisheye.cpp
HazarathKumarM: If dist is out of range, locSrc is already set to -1. Since nearest-neighbor interpolation checks ROI bounds and returns 0 for out-of-range locations, the destination pixels become black. Added a comment at the top to clarify this.

src/modules/tensor/hip/kernel/fisheye.cpp
HazarathKumarM: Added a comment here also to clarify this.

rrawther (approved):

AryanSalmanpour (approved):

PR #666, opened 2026-01-27
Erode - HOST and HIP update

r-abishek:
Adds tensor style implementation for Erode HOST and updates to HIP implementation.
Adds relevant QA unit and performance tests.


src/modules/tensor/rppt_tensor_morphological_operations.cpp
```diff
@@ -25,11 +25,82 @@ SOFTWARE.
...
+        erode_char_host_tensor(static_cast<Rpp8s*>(srcPtr) + srcDescPtr->offsetInBytes,
+                               srcDescPtr,
+                               static_cast<Rpp8s*>(dstPtr) + dstDescPtr->offsetInBytes,
+                               dstDescPtr,
+                               kernelSize,
+                               roiTensorPtrSrc,
+                               roiType,
+                               layoutParams,
+                               rpp::deref(rppHandle));
+    }
```
AryanSalmanpour, line 99: Add an else here and return a proper error. 

If srcDescPtr->dataType is not U8, F16, F32, or I8 (e.g., a future type), none of the if/else if branches run, and the function falls through, returning RPP_SUCCESS without writing to dstPtr. Callers would get success with an uninitialized/unchanged destination.

AryanSalmanpour (changes requested):

src/include/common/cpu/rpp_cpu_filter.hpp
```diff
@@ -694,4 +852,77 @@ inline void rpp_load_box_filter_float_9x9_host(__m256 *pRow, Rpp16f **srcPtrTemp
...
+    static inline VecType load(void *ptr) { return _mm256_cvtph_ps(_mm_castps_si128(_mm_loadu_ps((float*)ptr))); }
+};
+
+struct MorphPad_Erode
+{
+    static inline __m256i pad_int() { return _mm256_set1_epi8((char)255); }
+    static inline __m256  pad_float() { return avx_p1; }
+};
+
+// MorphPad_Dilate is defined for future morphological dilate implementation.
```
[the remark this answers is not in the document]
HazarathKumarM: Done

src/modules/tensor/rppt_tensor_morphological_operations.cpp
HazarathKumarM: Added else to return error

AryanSalmanpour (approved):

LakshmiKumar23 (approved):

PR #671, opened 2026-02-11
RPP Unified API for HIP/HOST

r-abishek:
This PR unifies HIP and HOST API for all RPP, and does all necessary support-changes for that.

**NOTE:**
* This PR changes RPP version to 3.1.0.
* This PR will need to be merged only in sync with an upcoming MIVisionX PR.

Changes:
* Changes handle to add a member of type RppBackend indicating backend for which the handle is created with rppCreate() function.
* Introduces an additional parameter to the external API indicating the backend for which augmentation is to be executed at runtime.
* Unification of the external API functions for HOST and HIP versions for all augmentations, enclosing the respective kernel calls inside conditional blocks based on user provided execution backend – Headers also updated with appropriate signatures.
* Updates all the external API calls across the test suite to use the unified API functions.
* Adds additional layout checks for image augmentations within the unified API, ensuring more robust check of user provided parameters.


LakshmiKumar23:
@r-abishek please resolve merge conflicts. Also, add a CHANGELOG since this is breaking API with vx_rpp. Please mention which versions are compatible with which vx_rpp versions

src/modules/tensor/rppt_tensor_audio_augmentations.cpp
```diff
@@ -35,570 +35,528 @@ SOFTWARE.
...
-                                    rpp::deref(rppHandle));
-    }
-    else
+    rpp::Handle &handle = rpp::deref(rppHandle);
+    RppBackend handleBackend = handle.GetBackend();
+
+    if(executionBackend == RppBackend::RPP_HOST_BACKEND)
     {
-        return RPP_ERROR_NOT_IMPLEMENTED;
+        // Disabled this check for now.
```
LakshmiKumar23, line 428: why are some of them disabled and others not?

LakshmiKumar23 (approved):

PR #672, opened 2026-02-12
[CMake]: Replace CMAKE_SOURCE_DIR -> RPP_SOURCE_DIR

zacharyvincze:
## PR Description:

* Replaces `CMAKE_SOURCE_DIR` with `RPP_SOURCE_DIR` to properly refer to the relative source directory for RPP.


zacharyvincze:
For future reference in this project. Please avoid using variables such as `CMAKE_SOURCE_DIR` or `CMAKE_BINARY_DIR` as this may break installs whenever RPP is being added to a larger project as a subdirectory.

Prefer to use the relative variants: `CMAKE_CURRENT_SOURCE_DIR` or `CMAKE_CURRENT_BINARY_DIR` instead. In the case of this, a variable is set in the root CMake: `RPP_SOURCE_DIR` to always refer to the root source directory of RPP, regardless of the location of any parent CMakeLists.txt files.

LakshmiKumar23 (approved):

kiritigowda (approved):

PR #681, opened 2026-02-27
Median and Gaussian Filter optimizations

r-abishek:
This PR contains specific powerful optimizations in Median and Gaussian filter to match and marginally beat AVX2 level performance using OpenCV.

PERFORMANCE COMPARISONS FOR MEDIAN FILTER AND GAUSSIAN FILTER
| image  | image |
| ------------- |:-------------:|
|image| image |


src/modules/tensor/cpu/kernel/median_filter.cpp
```diff
@@ -24,42 +24,1262 @@ SOFTWARE.
...
 x  x  x  x  x  x  x  x  x  x  ..  x  x
 x  1  2  3  4  5  6  7  8  9  .. 32  x
 x  1  2  3  4  5  6  7  8  9  .. 32  x
 x  1  2  3  4  5  6  7  8  9  .. 32  x
 x  x  x  x  x  x  x  x  x  x  ..  x  x
 
-padLength = 1 (kernelSize / 2)
+For each output pixel:
+1. Load 3×3 window with border replication (nearest-neighbor padding)
+2. Apply sorting network (25 compare-swap operations)
```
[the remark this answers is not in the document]
HazarathKumarM: Done

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
```diff
@@ -82,7 +82,25 @@ RppStatus gaussian_filter_host_tensor(T *srcPtr,
 #if __AVX2__
     __m256i pxMaskPln[7] = {avx_pxMaskRotate0To1, avx_pxMaskRotate0To2, avx_pxMaskRotate0To3, avx_pxMaskRotate0To4, avx_pxMaskRotate0To5, avx_pxMaskRotate0To6, avx_pxMaskRotate0To7};
     __m256i pxMaskPkd[7] = {avx_pxMaskRotate0To3, avx_pxMaskRotate0To6, avx_pxMaskRotate0To1, avx_pxMaskRotate0To4, avx_pxMaskRotate0To7, avx_pxMaskRotate0To2, avx_pxMaskRotate0To5};
+    
+    // Pre-allocate and broadcast filter coefficients for all batches before parallel loop
+    // Maximum kernel size is 9x9 = 81 coefficients
+    constexpr int MAX_FILTER_SIZE = 81;
+    __m256 *pFilterBatch = (__m256 *)aligned_alloc(32, dstDescPtr->n * MAX_FILTER_SIZE * sizeof(__m256));
```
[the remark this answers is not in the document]
HazarathKumarM: Added

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
```diff
@@ -1193,6 +1219,12 @@ RppStatus gaussian_filter_host_tensor(T *srcPtr,
             }
         }
     }
+    
+#if __AVX2__
+    // Free the pre-allocated filter memory
+    free(pFilterBatch);
+#endif
```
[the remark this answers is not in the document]
HazarathKumarM: Done

src/modules/tensor/hip/kernel/median_filter.cpp
```diff
@@ -300,135 +300,275 @@ __device__ void median_filter_3x3_row_hip_compute<float>(float* src_smem, d_floa
...
-        float3 val_f3 = make_float3(static_cast<float>(window[leftIdx]), static_cast<float>(window[midIdx]), static_cast<float>(window[rightIdx]));
-        float midVal = rpp_hip_median3(val_f3);
-
-        // 2. partition
-        int i = leftIdx;
-        int j = rightIdx;
-        while (i <= j)
+        // Build histogram (256 bins for U8)
+        int hist[256];
+        #pragma unroll
```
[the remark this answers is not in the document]
HazarathKumarM: Done

src/modules/tensor/hip/kernel/median_filter.cpp
```diff
@@ -300,135 +300,275 @@ __device__ void median_filter_3x3_row_hip_compute<float>(float* src_smem, d_floa
...
+
+                if (i <= j)
+                {
+                    T tmp = window[i];
+                    window[i] = window[j];
+                    window[j] = tmp;
+                    ++i;
+                    --j;
+                }
             }
```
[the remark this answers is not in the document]
HazarathKumarM: Done

src/modules/tensor/cpu/kernel/median_filter.cpp
```diff
@@ -24,42 +24,1262 @@ SOFTWARE.
...
+━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+1. Small kernels (3×3, 5×5): Sorting networks + AVX2 SIMD vectorization
+   - Uses hardcoded optimal compare-swap sequences
+   - Processes 8-32 elements simultaneously with SIMD
+   - No branching in inner loops
+
+2. Large kernels (7×7, 9×9+): Constant-time O(1) histogram method (U8 only)
+   - Two-tier histogram: coarse (4 MSB) + fine (full 8 bits)
+   - Maintains column histograms, slides window horizontally
+   - O(1) median finding regardless of kernel size
```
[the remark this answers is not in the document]
HazarathKumarM: Done

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
```diff
@@ -82,7 +82,25 @@ RppStatus gaussian_filter_host_tensor(T *srcPtr,
...
+    int filterSize = kernelSize * kernelSize;
+    for(int batchCount = 0; batchCount < dstDescPtr->n; batchCount++)
+    {
+        Rpp32f *filterTensor = handle.GetInitHandle()->mem.mcpu.scratchBufferHost + batchCount * kernelSize * kernelSize;
+        create_gaussian_kernel_host(filterTensor, stdDevTensor[batchCount], kernelSize);
+        
+        __m256 *pFilter = pFilterBatch + batchCount * MAX_FILTER_SIZE;
+        for (int i = 0; i < filterSize; i++)
+            pFilter[i] = _mm256_set1_ps(filterTensor[i]);
+    }
```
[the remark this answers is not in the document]
HazarathKumarM: Done

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
```diff
@@ -145,18 +160,29 @@ RppStatus gaussian_filter_host_tensor(T *srcPtr,
                         dstPtrTemp += padLength;
 #if __AVX2__
                         Rpp32s padIndex = (padVertical == RpptImageBorderEdge::BOTTOM_EDGE) ?  rowKernelLoopLimit - 1 : 0;
+                        // Prefetch next row for better cache performance
+                        if (i + 1 < roi.xywhROI.roiHeight)
+                        {
+                            for (int k = 0; k < kernelSize; k++)
```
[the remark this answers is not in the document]
HazarathKumarM: DOne

src/modules/tensor/cpu/kernel/median_filter.cpp
```diff
@@ -88,31 +1304,26 @@ inline void median_filter_generic_tensor(T *srcPtrTemp, T *dstPtrTemp, Rpp32s ro
 
     for (Rpp32s ch = 0; ch < channels; ch++)
     {
-        // Temporary buffer for the current channel's data in the kernel window
         T channelBlock[kernelSizeSquared];
```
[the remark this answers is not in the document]
HazarathKumarM: Done

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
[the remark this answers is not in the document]
rrawther: It is still computing outside OpenMP block.

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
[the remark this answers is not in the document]
rrawther: What does this mean? It is not resolved

src/modules/tensor/hip/kernel/median_filter.cpp
```diff
@@ -300,135 +300,286 @@ __device__ void median_filter_3x3_row_hip_compute<float>(float* src_smem, d_floa
...
-        int midIdx   = (leftIdx + rightIdx) >> 1;
-        float3 val_f3 = make_float3(static_cast<float>(window[leftIdx]), static_cast<float>(window[midIdx]), static_cast<float>(window[rightIdx]));
-        float midVal = rpp_hip_median3(val_f3);
-
-        // 2. partition
-        int i = leftIdx;
-        int j = rightIdx;
-        while (i <= j)
+        // Build histogram (256 bins for U8)
+        int hist[256];
```
AryanSalmanpour: A 256-element int array (1024 bytes) in registers per thread is not a good idea. The compiler will likely spill most of this to local memory, killing performance. 

Recommendation: 
Use shared memory for the histogram (one per thread or per-warp with atomics)

src/modules/tensor/hip/kernel/median_filter.cpp
```diff
@@ -300,135 +300,286 @@ __device__ void median_filter_3x3_row_hip_compute<float>(float* src_smem, d_floa
...
+        
+        // Unpack from row4
+        p[20] = static_cast<float>(src_smem[4 * SMEM_LENGTH_X + baseOffset]);
+        p[21] = static_cast<float>(src_smem[4 * SMEM_LENGTH_X + baseOffset + 1]);
+        p[22] = static_cast<float>(src_smem[4 * SMEM_LENGTH_X + baseOffset + 2]);
+        p[23] = static_cast<float>(src_smem[4 * SMEM_LENGTH_X + baseOffset + 3]);
+        p[24] = static_cast<float>(src_smem[4 * SMEM_LENGTH_X + baseOffset + 4]);
+
+        // Apply sorting network
+        #define OP(a, b) { float t = min(p[a], p[b]); p[b] = max(p[a], p[b]); p[a] = t; }
```
AryanSalmanpour: The exact same 90-operation sorting network appears twice (here and line 402 below)— once for the template <typename T> version and once for the template <> float specialization. The only difference is how p[] is loaded (cast from T vs direct float read).

Recommendation: Extract the sorting network into a `__device__` inline helper that takes `float p[25]` and returns the median. Both templates call it after loading.

src/modules/tensor/hip/kernel/median_filter.cpp
```diff
@@ -300,135 +300,286 @@ __device__ void median_filter_3x3_row_hip_compute<float>(float* src_smem, d_floa
     }
 }
 
+// Optimized 5×5 median using sorting network (processes 8 pixels simultaneously)
+template <typename T>
+__device__ void median_filter_5x5_row_hip_compute(T* src_smem, d_float8* median_f8)
+{
+    // Process 8 pixels using optimized sorting network
+    #pragma unroll
+    for (int px = 0; px < 8; ++px)
```
AryanSalmanpour, line 338: Unrolling the outer 8-pixel loop means the compiler will try to instantiate 8 copies of the 25-element sorting network simultaneously. That's 200 registers just for the p[] arrays (8 × 25 floats). Combined with the shared memory loads, this will cause massive register spill.

Recommendation: Remove #pragma unroll from this loop and let the compiler decide, or use #pragma unroll 1 to force no unrolling. The loop body is already substantial enough that unrolling hurts more than it helps.

src/modules/tensor/cpu/kernel/median_filter.cpp
```diff
@@ -24,43 +24,1268 @@ SOFTWARE.
...
+#undef op
+    return p[4];
+}
 
-4. Assign this median to the output pixel at (i, j)
+// 5x5 sorting network. After ops, p[12] is median.
+template <typename T>
+inline T rpp_median_5x5_sortnet(T *p)
+{
+#define op(i, j) rpp_minmax_op_scalar(p[i], p[j])
```
rrawther, line 92: I wonder if this performs better than C code with so many avx ops. Do you have a perf comparison. Simpler version given by AI is the following

rrawther: #include <immintrin.h>
#include <algorithm>
#include <cstdint>

// ============================================================================
// AVX2-optimized 5x5 median filter for signed char (int8_t) values
//
// Uses a bitonic sorting network on 25 elements to find the median.
// Processes 32 output pixels per iteration by operating across SIMD lanes.
// Each lane independently sorts its own 5x5 window and extracts the median.
// ============================================================================

// Scalar fallback for border pixels (nearest-neighbor clamping)
static inline char median_5x5_scalar(const char *src, int width, int height,
                                     int row, int col)
{
    char window[25];
    int idx = 0;
    for (int dy = -2; dy <= 2; dy++)
    {
        int r = std::max(0, std::min(row + dy, height - 1));
        for (int dx = -2; dx <= 2; dx++)
        {
            int c = std::max(0, std::min(col + dx, width - 1));
            window[idx++] = src[r * width + c];
        }
    }
    std::nth_element(window, window + 12, window + 25);
    return window[12];
}

#if __AVX2__
// Bitonic compare-and-swap: sort pair in ascending order
static inline void cas_asc(__m256i &a, __m256i &b)
{
    __m256i mn = _mm256_min_epi8(a, b);
    __m256i mx = _mm256_max_epi8(a, b);
    a = mn;
    b = mx;
}

// Bitonic compare-and-swap: sort pair in descending order
static inline void cas_desc(__m256i &a, __m256i &b)
{
    __m256i mn = _mm256_min_epi8(a, b);
    __m256i mx = _mm256_max_epi8(a, b);
    a = mx;
    b = mn;
}

// Bitonic sort 32 elements, extract median at index 12
// Elements 25-31 are padded with INT8_MAX so they sort to the top,
// leaving positions 0-24 holding the sorted real values.
static inline __m256i bitonic_sort_median25(__m256i r[25])
{
    __m256i rr[32];
    for (int i = 0; i < 25; i++) rr[i] = r[i];
    __m256i padVal = _mm256_set1_epi8(127);
    for (int i = 25; i < 32; i++) rr[i] = padVal;

    // Bitonic sort network for n=32 (5 stages)
    // Stage k: build bitonic sequences of length k
    // Substage j: compare-and-swap at distance j
    for (int k = 2; k <= 32; k <<= 1)
    {
        for (int j = k >> 1; j > 0; j >>= 1)
        {
            for (int i = 0; i < 32; i++)
            {
                int partner = i ^ j;
                if (partner > i)
                {
                    if ((i & k) == 0)
                        cas_asc(rr[i], rr[partner]);
                    else
                        cas_desc(rr[i], rr[partner]);
                }
            }
        }
    }
    return rr[12]; // median of 25 elements
}
#endif // __AVX2__

// Main entry point: 5x5 median filter for signed char images
// src: input image (row-major, single channel)
// dst: output image (same dimensions)
// width, height: image dimensions
void median_filter_5x5_char_avx2(const char *src, char *dst,
                                 int width, int height)
{
    const int padLen = 2;

    // Process all rows
    for (int row = 0; row < height; row++)
    {
        int col = 0;
        bool inVerticalBorder = (row < padLen || row >= height - padLen);

#if __AVX2__
        // AVX2 path: process 32 interior pixels at a time
        if (!inVerticalBorder)
        {
            // Scalar left border
            for (; col < padLen; col++)
                dst[row * width + col] = median_5x5_scalar(
                    src, width, height, row, col);

            // SIMD interior: 32 pixels per iteration
            for (; col + 32 <= width - padLen; col += 32)
            {
                __m256i r[25];
                int idx = 0;

                // Load 5 rows x 5 column offsets = 25 registers
                // Each register holds 32 pixels from consecutive columns,
                // shifted by dx relative to the output column position
                for (int dy = -2; dy <= 2; dy++)
                {
                    const char *rowPtr = src + (row + dy) * width + col;
                    for (int dx = -2; dx <= 2; dx++)
                        r[idx++] = _mm256_loadu_si256(
                            (const __m256i *)(rowPtr + dx));
                }

                // Sort and extract median
                __m256i median = bitonic_sort_median25(r);

                // Store 32 output pixels
                _mm256_storeu_si256(
                    (__m256i *)(dst + row * width + col), median);
            }

            // Scalar right border + remainder
            for (; col < width; col++)
                dst[row * width + col] = median_5x5_scalar(
                    src, width, height, row, col);
            continue;
        }
#endif // __AVX2__

        // Scalar fallback for border rows (or when AVX2 is unavailable)
        for (; col < width; col++)
            dst[row * width + col] = median_5x5_scalar(
                src, width, height, row, col);
    }
}

rrawther (changes requested):

src/modules/tensor/cpu/kernel/median_filter.cpp
HazarathKumarM: Currently we are already using AVX2 optimized sorting network for the 5×5 median filter. we use the same SIMD min/max compare-swap pattern.
We process 32 pixels at once for U8/I8 types and 8 pixels for F32/F16, using the same idea you mentioned.

One difference is, instead of using generic bitonic sort (which needs padding 25 -> 32 elements and require around 80 compare-swap operations), we use a median network for 25 elements with around 60 compare-swap operations. So this reduces the number of comparisons while still keeping the same SIMD benefits.

src/modules/tensor/hip/kernel/median_filter.cpp
HazarathKumarM: Done

src/modules/tensor/hip/kernel/median_filter.cpp
HazarathKumarM: Done

src/modules/tensor/hip/kernel/median_filter.cpp
HazarathKumarM: Removed

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
```diff
@@ -54,10 +54,10 @@ inline void create_gaussian_kernel_host(Rpp32f* filter, Rpp32f stdDev, int kerne
     Rpp32f kernelSum = 0.0f;
     for (int i = 0; i < kernelSize * kernelSize; i++)
         kernelSum += filter[i];
-    kernelSum = 1.0f / kernelSum;
-
+    
+    Rpp32f invSum = 1.0f / kernelSum;
     for (int i = 0; i < kernelSize * kernelSize; i++)
-        filter[i] *= kernelSum;
+        filter[i] *= invSum;
```
rrawther: This is another suggestion for more efficient kernel generation by combining normalization with kernel generation
Rpp32f kernelSum = 0.0f;
for (int i = -kernelHalfSize; i <= 0; i++, rowIdx += kernelSize)
{
    int iSquare = i * i;
    Rpp32f rowSum = 0.0f;
    for (int j = -kernelHalfSize; j <= 0; j++)
    {
        Rpp32f val = gaussian(iSquare, j, mulFactor);
        filter[rowIdx + (kernelHalfSize + j)] = filter[rowIdx + (kernelHalfSize - j)] = val;
        rowSum += (j == 0) ? val : 2.0f * val;  // center once, mirrored pair twice
    }
    // If this row was memcpy'd to a mirrored row, count it twice
    bool mirrored = ((kernelSize * (kernelSize - 1) - rowIdx) != rowIdx);
    kernelSum += mirrored ? 2.0f * rowSum : rowSum;
    if (mirrored)
        std::memcpy(&filter[kernelSize * (kernelSize - 1) - rowIdx], &filter[rowIdx], kernelSize * sizeof(float));
}

// Normalize — only one loop remains
Rpp32f invSum = 1.0f / kernelSum;
for (int i = 0; i < kernelSize * kernelSize; i++)
    filter[i] *= invSum;

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
```diff
@@ -82,7 +82,9 @@ RppStatus gaussian_filter_host_tensor(T *srcPtr,
 #if __AVX2__
     __m256i pxMaskPln[7] = {avx_pxMaskRotate0To1, avx_pxMaskRotate0To2, avx_pxMaskRotate0To3, avx_pxMaskRotate0To4, avx_pxMaskRotate0To5, avx_pxMaskRotate0To6, avx_pxMaskRotate0To7};
     __m256i pxMaskPkd[7] = {avx_pxMaskRotate0To3, avx_pxMaskRotate0To6, avx_pxMaskRotate0To1, avx_pxMaskRotate0To4, avx_pxMaskRotate0To7, avx_pxMaskRotate0To2, avx_pxMaskRotate0To5};
+    constexpr int MAX_FILTER_SIZE = 81;  // Maximum kernel size is 9x9 = 81 coefficients
```
rrawther: please use a #define macro for this instead

src/modules/tensor/cpu/kernel/median_filter.cpp
```diff
@@ -135,38 +1352,227 @@ RppStatus median_filter_generic_host_tensor(T *srcPtr,
...
+                    {
+                        // Scalar fallback
+                        T *dstPtrRow = dstPtrChannel;
+                        for(Rpp32s i = 0; i < roi.xywhROI.roiHeight; i++)
+                        {
+                            T *dstPtrTemp = dstPtrRow;
+                            for(Rpp32s j = 0; j < roi.xywhROI.roiWidth; j++)
+                            {
+                                if (useSortNet3)
+                                    median_filter_3x3_sortnet_tensor(srcPtrChannel, dstPtrTemp, i, j, roi.xywhROI.roiHeight - 1, roi.xywhROI.roiWidth - 1, 1, srcDescPtr);
```
rrawther: i don't think we should post_fix all names with tensor (median_filter_3x3_sortnet would suffice)

src/modules/tensor/cpu/kernel/median_filter.cpp
```diff
@@ -135,38 +1352,227 @@ RppStatus median_filter_generic_host_tensor(T *srcPtr,
...
+                                {
+                                    const Rpp8u *r0 = (const Rpp8u *)srcPtrChannel + std::max(i - 2, 0) * srcDescPtr->strides.hStride;
+                                    const Rpp8u *r1 = (const Rpp8u *)srcPtrChannel + std::max(i - 1, 0) * srcDescPtr->strides.hStride;
+                                    const Rpp8u *r2 = (const Rpp8u *)srcPtrChannel + i * srcDescPtr->strides.hStride;
+                                    const Rpp8u *r3 = (const Rpp8u *)srcPtrChannel + std::min(i + 1, roi.xywhROI.roiHeight - 1) * srcDescPtr->strides.hStride;
+                                    const Rpp8u *r4 = (const Rpp8u *)srcPtrChannel + std::min(i + 2, roi.xywhROI.roiHeight - 1) * srcDescPtr->strides.hStride;
+                                    rpp_median5x5_pln_u8_avx(r0, r1, r2, r3, r4, dstRow, roi.xywhROI.roiWidth);
+                                }
+                            }
+                            useAvx = true;
```
rrawther: why we are hardcoding useAvx=true here and false before

rrawther (commented):

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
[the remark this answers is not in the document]
HazarathKumarM: The code has been changed to use stack allocated pFilter[MAX_FILTER_SIZE] inside the parallel loop instead of heap allocation with aligned_alloc(). This eliminates the memory leak risk as stack memory is automatically cleaned up

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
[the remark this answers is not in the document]
HazarathKumarM: The filter computation (create_gaussian_kernel_host and the AVX broadcast loop) has been moved inside the #pragma omp parallel for  block

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
HazarathKumarM: Done. Updated the code with the suggestion

src/modules/tensor/cpu/kernel/gaussian_filter.cpp
HazarathKumarM: Done

src/modules/tensor/cpu/kernel/median_filter.cpp
HazarathKumarM: Done. Renamed all the instances

src/modules/tensor/cpu/kernel/median_filter.cpp
HazarathKumarM: Removed the bool avxProcessed variables and changed to use if-else pattern with #if __AVX2__ / #endif guards

LakshmiKumar23 (approved):

LakshmiKumar23:
@r-abishek @HazarathKumarM code coverage drop 7% . Please take a look

HazarathKumarM:
@LakshmiKumar23 With the recent CI runs, we are observing the code coverage of this PR as 92.48%. we are not observing any drop
image

LakshmiKumar23 (approved):

rrawther (approved):

PR #694, opened 2026-03-24
Fix - Address RPP -Wall warnings (Phase 2)

zacharyvincze:
## Description
Removes remaining `-Wall` warnings from RPP's codebase. Since this changes many files, it will require a more in-depth review to ensure that no correctness errors arise.

Note that `-Wvla-cxx-extension` is being suppressed, since this must be compiled with a Clang compiler regardless. `amdclang++` supports variable length arrays, even though this isn't a C++ standard.

This branch is based on `zv/bugfix/resolve-warnings-phase-1`, and must be rebased once that PR is merged into develop. This is why the diff (compared to develop) is very large. Keeping this as a draft PR for now until that branch gets merged in.

## Testing
Standard tests, via `ctest -VV` are all passing with these changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,19 +152,11 @@ endif()
 # Enable -Wall warnings for both release and debug builds
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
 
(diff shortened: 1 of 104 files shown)


LakshmiKumar23 (approved):
Passes Ctests on local system

PR #712, opened 2026-04-17
[WIP] Bug Fix -  jpeg compression distortion - HIP and HOST

LakshmiKumar23:
## Motivation

There are 2 failures on ctest (not reproducible on Navi31 and MI325 locally). Claude suggested fixes 

## Technical Details

Root causes addressed for `jpeg_compression_distortion_f32_Tensor_PLN1_to_PLN1`

- `HIP FLOAT4_ONE_OVER_255` vs CPU `ONE_OVER_255`

HIP used a separate literal (`0.003921569f`) while HOST uses `0.00392156862745f`. As float they often match, but defining HIP from `ONE_OVER_255` keeps the path identical and avoids drift.

- Scalar `quantize_block` vs AVX `quantize_block_avx2`

The scalar path used `roundf` while AVX uses `std::round` in `accurate_quant_round`. That can disagree on some quotients and shows up as strict QA failures. Scalar quantization now uses the same `std::round(quotient)` pattern (with `#include <cmath>`).

- F32 PLN1 aligned loop without AVX2

When `__AVX2__` is off, the aligned for still advanced `srcPtrTemp / dstPtrTemp` by 8 but never ran the AVX block, so most columns were never processed. An #else branch now calls` jpeg_compression_distortion_pln1_generic` for that strip (same idea as the width tail). This matters for any CI or reference build compiled without AVX2.

- HIP quantize

Still uses `roundf` on device (portable in` __device__` code). The comment was updated; behavior should match HOST after (2).

Files touched
`src/include/common/hip/rpp_hip_load_store.hpp` — `FLOAT4_ONE_OVER_255` from `ONE_OVER_255`
`src/modules/tensor/cpu/kernel/jpeg_compression_distortion.cpp` — `<cmath>, quantize_block` rounding, F32 PLN1 non-AVX2 path
`src/modules/tensor/hip/kernel/jpeg_compression_distortion.cpp` — small clarification in quantize (quotient variable)

## Test Plan

Run ctest for imageHost and imageHIP

## Test Result

all jpeg compression distortion must pass

## Submission Checklist

### DIFF
--- a/src/include/common/hip/rpp_hip_load_store.hpp
+++ b/src/include/common/hip/rpp_hip_load_store.hpp
@@ -124,7 +124,7 @@ struct RoundToNearest {};
 #define SMEM_LENGTH_X                   128                 // Shared memory length of 128 cols to efficiently utilize all 16 LOCAL_THREADS_X as 16 * 8-byte vectorized global read/writes per thread = 128 bytes, fitting in 32 banks 4 byte wide
 #define SMEM_LENGTH_Y_1C                16                  // Shared memory length of 16 rows to efficiently utilize all 16 LOCAL_THREADS_Y as 1 128-byte-long row per thread (single channel greyscale)
 #define SMEM_LENGTH_Y_3C                48                  // Shared memory length of 48 rows to efficiently utilize all 16 LOCAL_THREADS_Y as 3 128-byte-long rows per thread (three channel rgb)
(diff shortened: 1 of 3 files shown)


LakshmiKumar23:
Fixed with change in build type in CI: https://github.com/ROCm/rocJenkins/pull/1747

PR #713, opened 2026-04-20
[Cleanup] Remove stdout/stderr prints in core library implementations

### LINKED ISSUE #698: Remove "printf" from RPP core library
printfs are not recommended for libraries. 
Need to replace with proper Error logging mechanism

zacharyvincze:
# Description
This PR aims to fully remove prints to stdout/stderr from the core library implementation, instead opting for error propagation through the standard RppStatus return codes.

# Technical Details
- Resolves #698
- Fixes bugs in kernel dispatchers where RppStatus was not being properly propagated from the kernel call, causing some errors to silently fail.
- Fixes some branching errors where unsupported combinations of parameters would silently fall through to an RPP_SUCCESS status without running anything.
- SIMD print helpers now only print when `RPP_DEBUG_SIMD_PRINTS` is set as a build flag to avoid printing in the core library. These should be enabled when debugging is required.

# Known Issues
- Checks for HIP runtime calls in RPP kernel dispatchers lose information about the HIP error that occurred since we can only return `RppStatus`. This may be too generic for debugging purposes. Previously, the HIP error was printed out before calling `exit(-1)`. Hard crashes such as these cannot occur in a library which is expected to propagate its error and let the user handle it accordingly. We will need to come back to this later and offer a better solution.


src/include/common/cpu/rpp_cpu_simd_load_store.hpp
```diff
@@ -195,15 +195,19 @@ union RppSIMD256
     float    f32[8];
 };
 
-// Print helpers
+// Print helpers (compiled only when RPP_DEBUG_SIMD_PRINTS is defined)
 
 inline void rpp_mm_print_epi8(__m128i *v)
 {
+#if defined(RPP_DEBUG_SIMD_PRINTS)
```
LakshmiKumar23: @zacharyvincze where is this being defined. I only see all `#if defined()` in this PR. Is this to be passed as `-D` flag for Cmake?

LakshmiKumar23 (approved):

src/include/common/errors.hpp
```diff
@@ -84,14 +83,13 @@ rppStatus_t try_(F f, bool output = true)
     }
     catch(const Exception& ex)
     {
-        if(output)
-            std::cerr << "RPP Error: " << ex.what() << std::endl;
+        (void)output;
```
rrawther: The output parameter is now completely ignored. If it was meaningful API, this is a silent behavioral change. Consider deprecating the parameter or removing it entirely rather than silencing it with (void).

src/include/common/errors.hpp
```diff
@@ -84,14 +83,13 @@ rppStatus_t try_(F f, bool output = true)
...
         return ex.status;
     }
     catch(const std::exception& ex)
     {
-        if(output)
-            std::cerr << "RPP Error: " << ex.what() << std::endl;
+        (void)output;
+        (void)ex;
```
rrawther: why are we discarding exception here

src/modules/tensor/cpu/kernel/normalize.cpp
```diff
@@ -680,7 +682,8 @@ RppStatus normalize_f32_f32_host_tensor(Rpp32f *srcPtr,
                     break;
                 }
                 default:
-                    std::cout<<"Invalid Axis mask"<<std::endl;
+                    axisMaskStatus.store(RPP_ERROR_INVALID_AXIS, std::memory_order_relaxed);
```
rrawther, line 685: Thread-safety of std::atomic error reporting in OpenMP loops - The pattern using std::atomic<RppStatus> with memory_order_relaxed + continue is correct for detecting any error, but there's a subtlety: when an incompatible dimension is found for one sample in the batch, other threads continue processing. This means partial results may be written before the error is returned. Consider whether this is acceptable or if you should skip the write for failed samples.

Suggested by claude. Please check if this happens

api/rppdefs.h
```diff
@@ -64,24 +64,26 @@ typedef halfhpp Rpp16f;
...
+    hipError_t _rpp_hip_err = (expr);                                          \
+    if (_rpp_hip_err != hipSuccess)                                            \
+    {                                                                          \
+        fprintf(stderr, "HIP runtime error: %s returned %d at %s:%d", #expr, _rpp_hip_err, __FILE__, __LINE__); \
+        return RPP_ERROR_HIP_RUNTIME;                                          \
+    }                                                                          \
+  } while (0)
+
 /*! \brief Check last HIP error after kernel launch; return RPP_ERROR_HIP_LAUNCH on failure. Use after hipLaunchKernelGGL. \ingroup group_rppdefs */
 #define HIP_CHECK_LAUNCH_RETURN()                                              \
```
rrawther: The std::cerr print was removed from HIP_CHECK_LAUNCH_RETURN but unlike RPP_HIP_RETURN_IF_ERROR, it doesn't print the error to stderr. For consistency, consider adding a fprintf(stderr, ...) here too.

src/modules/tensor/hip/kernel/color_space_conversion.cpp
```diff
@@ -143,9 +143,9 @@ static void rpp_nv12_set_mat_yuv2rgb(RpptColorStandard col_standard, RpptColorRa
         for (int j = 0; j < 3; j++)
             mat[i][j] = (float)(1.0 * max_val / (white - black) * mat[i][j]);
     hipError_t status = hipMemcpyToSymbol(rpp_nv12_yuv_to_rgb_mat, mat, sizeof(mat));
-    CHECK_RETURN_STATUS(status);
-    status = hipMemcpyToSymbol(rpp_nv12_y_bias, &black, sizeof(black));
-    CHECK_RETURN_STATUS(status);
+    if(status != hipSuccess)
```
rrawther, line 154: why not add RPP_HIP_RETURN_IF_ERROR here? Also the below line

src/modules/tensor/hip/kernel/normalize.cpp
```diff
@@ -1552,9 +1558,10 @@ RppStatus hip_exec_compute_mean_stddev_tensor(T *srcPtr,
     if (tensorDims == 2)
     {
         // set the block and grid configuration based on axisMask
-        set_kernel_launch_config_2d(srcGenericDescPtr, globalThreads_x, globalThreads_y, globalThreads_z,
-                                    localThreads_x, localThreads_y, localThreads_z, axisMask,
-                                    partialSumArr, handle);
+        if (RppStatus cfgStatus = set_kernel_launch_config_2d(srcGenericDescPtr, globalThreads_x, globalThreads_y, globalThreads_z,
```
rrawther, line 1561: Can be combined as below
if ((auto cfgStatus = (set_kernel_launch_config_2d(srcGenericDescPtr, globalThreads_x, globalThreads_y, globalThreads_z,
                                                              localThreads_x, localThreads_y, localThreads_z, axisMask,
                                                              partialSumArr, handle))!= RPP_SUCCESS) {

                                                              }

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
```diff
@@ -567,19 +567,21 @@ RppStatus rppt_sobel_filter(RppPtr_t srcPtr,
...
-            CHECK_RETURN_STATUS(hipMalloc(&tempPtr, dataSize));
+            RPP_HIP_RETURN_IF_ERROR(hipMalloc(&tempPtr, dataSize));
 
             RpptSubpixelLayout srcSubpixelLayout = RpptSubpixelLayout::RGBtype;
             RppStatus errorStatus = rppt_color_to_greyscale(srcPtr, srcDescPtr, tempPtr, dstDescPtr, srcSubpixelLayout, rppHandle, RppBackend::RPP_HIP_BACKEND);        
             if(errorStatus != RPP_SUCCESS)
             {
-                CHECK_RETURN_STATUS(hipFree(tempPtr));
+                // Ignore the error status of hipFree to preserve the root cause of the error
+                auto status = hipFree(tempPtr);
```
rrawther: please remove status if not used

rrawther (changes requested):

src/modules/tensor/hip/kernel/color_space_conversion.cpp
zacharyvincze: RPP_HIP_RETURN_IF_ERROR returns an RppStatus, while the function signature for this expects a hipError_t.

src/include/common/errors.hpp
zacharyvincze: Regarding this and the comment above. I'll re-add the printouts assuming output is set to true. Given the nature of this PR/issue, can I make the suggestion to set the default value of `output` to false as to not print out error messages in the core library?

src/modules/tensor/hip/kernel/normalize.cpp
zacharyvincze: `auto cfgStatus = ...` is a declaration, not an assignment. So this change would produce a syntax error.

rrawther: YIK. But I believe the compiler will check for the assigned value if you add the braces. Worth trying if it is giving errors

rrawther:
@zacharyvincze  Few comments still need resolution

src/modules/tensor/cpu/kernel/normalize.cpp
zacharyvincze: I think this is acceptable for now (the user can discard the output if an error is thrown). But we can come back to it later. The validation steps for this operator are beyond the scope of this PR.

zacharyvincze:
@rrawther resolved the comments and synced with develop. I've re-added the printouts for internal HIP runtime errors (and HIP launch errors). Good to merge if everything looks good on your end.

rrawther (approved):

PR #716, opened 2026-05-01
[Bug Fix] Fix for the incorrect 8-bit chroma range for studio video range in YUV to RGB conversion matrix setup.

jeffqjiangNew:
## Motivation

This PR fixes the incorrect 8-bit chroma range for studio video range in YUV to RGB conversion matrix setup.

## Technical Details
For studio range YUV samples, the luma and chroma ranges are used to calculate the scaling factor:
  - We were using the luma range [16, 235] (span=219) for chroma.
  - Now we use the correct chroma range: [16, 240] (span=224).
  - Updated reference driving_frame_2048x1024.rgb with the version generated by OpenCV.

## Test Plan

Run test_rocm_interface_cpu_fallback test in TorchCodec with ROCm decoder enabled.

## Test Result

The test_rocm_interface_cpu_fallback should now pass.

## Submission Checklist


src/modules/tensor/hip/kernel/color_space_conversion.cpp
```diff
@@ -106,9 +106,12 @@ static void rpp_nv12_set_mat_yuv2rgb(RpptColorStandard col_standard, RpptColorRa
 {
     float wr = 0.2126f, wb = 0.0722f;
     int black = 16, white = 235, max_val = 255;
+    int chroma_min = 16, chroma_max = 240;
     if (color_range == RpptColorRange_FULL) {
         black = 0;
         white = 255;
+        chroma_min = 0;
```
rrawther: if the values are the same as black and white, we can reuse the same instead of introducing 2 new variables

src/modules/tensor/hip/kernel/color_space_conversion.cpp
```diff
@@ -139,9 +142,17 @@ static void rpp_nv12_set_mat_yuv2rgb(RpptColorStandard col_standard, RpptColorRa
...
         {1.0f, (1.0f - wb) / 0.5f, 0.0f},
     };
-    for (int i = 0; i < 3; i++)
-        for (int j = 0; j < 3; j++)
-            mat[i][j] = (float)(1.0 * max_val / (white - black) * mat[i][j]);
+    // Luma and chroma have different ranges in studio/TV mode:
+    //   Y:     [black, white]       = [16, 235] → span 219
+    //   Cb/Cr: [chroma_min, chroma_max] = [16, 240] → span 224
+    // Full range uses [0, 255] for both.
+    float y_scale = (float)max_val / (white - black);
```
rrawther: y_scale and c_scale are the same as per const. consider using only one scale value

rrawther (changes requested):

src/modules/tensor/hip/kernel/color_space_conversion.cpp
jeffqjiangNew: The code is simplified. Please take a look.

src/modules/tensor/hip/kernel/color_space_conversion.cpp
jeffqjiangNew: The code is simplified. Please take a look.

LakshmiKumar23 (approved):
Passes CI when compared with the OpenCV golden output

LakshmiKumar23:
CI running: https://math-ci.amd.com/blue/organizations/jenkins/mainline%2Fprecheckin%2Frpp/detail/PR-716/3/pipeline

rrawther (approved):

PR #717, opened 2026-05-14
[Bugfix] Fix use after free, synchronization issues, and incorrect RppStatus propagation in Sobel filter

### LINKED ISSUE #687: [Issue] Potential memory leak/use after free for tempPtr in Sobel HIP path
## Issue Description

The Sobel implementation contains issues with memory management related to the HIP path shown below:

https://github.com/ROCm/rpp/blob/187caaba49e0fa7be35862b161d6baf2e96c6d61/src/modules/tensor/rppt_tensor_filter_augmentations.cpp#L561-L640

### Resource Management Issues
* `tempPtr` is allocated using `hipMalloc` if the input tensor has 3 channels. If the branch which returns `RPP_ERROR_NOT_IMPLEMENTED` is taken, the function returns early without freeing the allocated memory.
* `hip_exec_sobel_filter_tensor` may take an RPP handle containing a stream reference, therefore launching asynchronously. Since there is no stream synchronization after this call, `hipFree` may be called before `hip_exec_soble_filter_tensor` is finished with its work. Leading to a use-after-free error.

### Status Propagation Issues
* `hip_exec_sobel_filter_tensor` returns an `RppStatus` which is not propagated to its caller: `rppt_sobel_filter`. This may hide potential error statuses which `hip_exec_soble_filter_tensor` would return. This issue also exists in the Host path for the Sobel filter implementation.

zacharyvincze:
## Technical Details
- Resolves #687 
- Use asynchronous versions for HIP runtime calls to honor stream semantics.
- Ensure return value for `sobel_filter` host/hip` kernels is checked and propagated by dispatcher.
- Ensure `tempPtr` is always cleaned up even when an error may cause an early return.

### DIFF
--- a/src/modules/tensor/rppt_tensor_filter_augmentations.cpp
+++ b/src/modules/tensor/rppt_tensor_filter_augmentations.cpp
@@ -506,56 +506,54 @@ RppStatus rppt_sobel_filter(RppPtr_t srcPtr,
 
         if ((srcDescPtr->dataType == RpptDataType::U8) && (dstDescPtr->dataType == RpptDataType::U8))
         {
-            sobel_filter_host_tensor(static_cast<Rpp8u*>(tempPtr) + srcDescPtr->offsetInBytes,
-                                     dstDescPtr,
-                                     static_cast<Rpp8u*>(dstPtr) + dstDescPtr->offsetInBytes,
-                                     dstDescPtr,
-                                     sobelType,
-                                     kernelSize,
-                                     roiTensorPtrSrc,
-                                     roiType,
-                                     handle);
+            return sobel_filter_host_tensor(static_cast<Rpp8u*>(tempPtr) + srcDescPtr->offsetInBytes,
+                                            dstDescPtr,
+                                            static_cast<Rpp8u*>(dstPtr) + dstDescPtr->offsetInBytes,
+                                            dstDescPtr,
+                                            sobelType,
+                                            kernelSize,
+                                            roiTensorPtrSrc,
+                                            roiType,
+                                            handle);
         }
         else if ((srcDescPtr->dataType == RpptDataType::F16) && (dstDescPtr->dataType == RpptDataType::F16))
         {
-            sobel_filter_host_tensor(reinterpret_cast<Rpp16f*>(static_cast<Rpp8u*>(tempPtr) + srcDescPtr->offsetInBytes),
-                                     dstDescPtr,
-                                     reinterpret_cast<Rpp16f*>(static_cast<Rpp8u*>(dstPtr) + dstDescPtr->offsetInBytes),
-                                     dstDescPtr,
-                                     sobelType,
-                                     kernelSize,
-                                     roiTensorPtrSrc,
-                                     roiType,
-                                     handle);
+            return sobel_filter_host_tensor(reinterpret_cast<Rpp16f*>(static_cast<Rpp8u*>(tempPtr) + srcDescPtr->offsetInBytes),
+                                            dstDescPtr,
+                                            reinterpret_cast<Rpp16f*>(static_cast<Rpp8u*>(dstPtr) + dstDescPtr->offsetInBytes),
+                                            dstDescPtr,
+                                            sobelType,
+                                            kernelSize,
+                                            roiTensorPtrSrc,
+                                            roiType,
+                                            handle);
         }
         else if ((srcDescPtr->dataType == RpptDataType::F32) && (dstDescPtr->dataType == RpptDataType::F32))
         {
-            sobel_filter_host_tensor(reinterpret_cast<Rpp32f*>(static_cast<Rpp8u*>(tempPtr) + srcDescPtr->offsetInBytes),
-                                     dstDescPtr,
-                                     reinterpret_cast<Rpp32f*>(static_cast<Rpp8u*>(dstPtr) + dstDescPtr->offsetInBytes),
-                                     dstDescPtr,
-                                     sobelType,
-                                     kernelSize,
-                                     roiTensorPtrSrc,
-                                     roiType,
-                                     handle);
+            return sobel_filter_host_tensor(reinterpret_cast<Rpp32f*>(static_cast<Rpp8u*>(tempPtr) + srcDescPtr->offsetInBytes),
+                                            dstDescPtr,
+                                            reinterpret_cast<Rpp32f*>(static_cast<Rpp8u*>(dstPtr) + dstDescPtr->offsetInBytes),
+                                            dstDescPtr,
+                                            sobelType,
+                                            kernelSize,
+                                            roiTensorPtrSrc,
+                                            roiType,
+                                            handle);
         }
         else if ((srcDescPtr->dataType == RpptDataType::I8) && (dstDescPtr->dataType == RpptDataType::I8))
         {
-            sobel_filter_host_tensor(static_cast<Rpp8s*>(tempPtr) + srcDescPtr->offsetInBytes,
-                                     dstDescPtr,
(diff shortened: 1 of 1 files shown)


src/modules/tensor/rppt_tensor_filter_augmentations.cpp
```diff
@@ -506,56 +506,54 @@ RppStatus rppt_sobel_filter(RppPtr_t srcPtr,
...
-                                     sobelType,
-                                     kernelSize,
-                                     roiTensorPtrSrc,
-                                     roiType,
-                                     handle);
+            return sobel_filter_host_tensor(static_cast<Rpp8u*>(tempPtr) + srcDescPtr->offsetInBytes,
+                                            dstDescPtr,
+                                            static_cast<Rpp8u*>(dstPtr) + dstDescPtr->offsetInBytes,
+                                            dstDescPtr,
+                                            sobelType,
```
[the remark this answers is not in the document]
zacharyvincze: Resolved this bug which existed on the HOST path.

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
```diff
@@ -567,80 +568,80 @@ RppStatus rppt_sobel_filter(RppPtr_t srcPtr,
...
-                // Ignore the error status of hipFree to preserve the root cause of the error
-                (void)hipFree(tempPtr);
+                // Ignore hipFree status to preserve the root cause of the error
+                (void)hipFreeAsync(tempPtr, handle.GetStream());
                 return errorStatus;
             }
             inputDesc = dstDescPtr;
         }
         srcPtr = (tempPtr == nullptr) ? srcPtr : tempPtr;
-        RPP_HIP_RETURN_IF_ERROR(hipStreamSynchronize(handle.GetStream()));
```
LakshmiKumar23, line 583: @zacharyvincze why was this removed. I remember @msiddaiah and I added this because there were some sync issues and the tests were randomly failing

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
```diff
@@ -567,80 +568,80 @@ RppStatus rppt_sobel_filter(RppPtr_t srcPtr,
...
+                                                  handle);
         }
         else
-            return RPP_ERROR_NOT_IMPLEMENTED;
+            status = RPP_ERROR_NOT_IMPLEMENTED;
 
         if (tempPtr != nullptr)
         {
-            // Sobel runs asynchronously on the same stream; wait before freeing the greyscale scratch buffer.
-            RPP_HIP_RETURN_IF_ERROR(hipStreamSynchronize(handle.GetStream()));
```
LakshmiKumar23, line 639: same as above

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
```diff
@@ -495,67 +495,68 @@ RppStatus rppt_sobel_filter(RppPtr_t srcPtr,
...
+        RpptDescPtr inputDesc = srcDescPtr;
         if (srcDescPtr->c == 3)
         {
             RpptSubpixelLayout srcSubpixelLayout = RpptSubpixelLayout::RGBtype;
             tempPtr = handle.GetInitHandle()->mem.mcpu.scratchBufferHost;
             RppStatus errorStatus = rppt_color_to_greyscale(srcPtr, srcDescPtr, tempPtr, dstDescPtr, srcSubpixelLayout, rppHandle, RppBackend::RPP_HOST_BACKEND);
             if(errorStatus != RPP_SUCCESS)
                 return errorStatus;
+            // Greyscale wrote to tempPtr using dstDescPtr's layout/offset; sobel must read with the same descriptor.
+            inputDesc = dstDescPtr;
```
rrawther, line 507: this has no effect since there is a return statement before. Please remove it

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
```diff
@@ -495,67 +495,68 @@ RppStatus rppt_sobel_filter(RppPtr_t srcPtr,
     {
         // convert image to grey scale if input is RGB image
         RppPtr_t tempPtr = srcPtr;
+        RpptDescPtr inputDesc = srcDescPtr;
```
rrawther, line 498: why is this required. Please use srcDescPtr directly in all places

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
```diff
@@ -495,67 +495,68 @@ RppStatus rppt_sobel_filter(RppPtr_t srcPtr,
...
-                                     dstDescPtr,
-                                     static_cast<Rpp8u*>(dstPtr) + dstDescPtr->offsetInBytes,
-                                     dstDescPtr,
-                                     sobelType,
-                                     kernelSize,
-                                     roiTensorPtrSrc,
-                                     roiType,
-                                     handle);
+            return sobel_filter_host_tensor(static_cast<Rpp8u*>(tempPtr) + inputDesc->offsetInBytes,
+                                            inputDesc,
```
rrawther, line 513: use srcDescPtr instead in all such

rrawther (changes requested):
Please address the review comments

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
zacharyvincze: Sobel internally converts an RGB image into a grayscale image for the sobel filter conditionally. `inputDesc` is meant to keep track of which one we're using in the actual sobel filter kernel. If we use `srcDescPtr` instead of `inputDesc` in the sobel filter kernel, and `srcDescPtr` is describing an RGB image, then we would be providing an incorrect descriptor.

This same logic applies to the similar comment you made regarding the use of `inputDesc`.

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
zacharyvincze: The return statement is only called when the branch above is taken (an error has occurred). So this is not dead code.

It is here to ensure that we are properly describing the converted grayscale image, rather than the RGB image that is passed in when taking this branch (`srcDescPtr->c == 3`).

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
zacharyvincze: Same as above.

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
zacharyvincze: Synchronization issues were mostly likely occurring since we were using `hipFree` and `hipMalloc` which both operate on the null stream, while the kernel itself was operating on the user provided stream. This means that it was possible for the `hipFree` to run (and free the grayscale buffer being used) before the kernel had finished executing, leading to a race condition.

To remedy this, we are now using the asynchronous versions of these runtime calls (`hipMallocAsync`, `hipFreeAsync`) which operate on the same user-provided stream that the kernels use.

As an aside, we shouldn't be using stream synchronization calls within the dispatcher implementation. The function signature leads the user to believe that this is a full asynchronous function that they are responsible for joining on. Blocking the host would go against that behavior.

rrawther: I agree with zach here

src/modules/tensor/rppt_tensor_filter_augmentations.cpp
rrawther: Agree. I think I missed the reassignment

rrawther (approved):

PR #726, opened 2026-06-04
Opencv rpp benchmark

msiddaiah:
## Motivation

Competitive analysis of RPP HOST operators and equivalent OPENCV implementations

## Technical Details

C++ Benchmarking infrastructure is built to measure OPENCV performance against RPP HOST operators under utilities/test_suite/OPENCV folder

## Test Plan

run the following shell script to get the benchmarking
utilities/test_suite/OPENCV /run_benchmarking.sh 

## Test Result

The above script will generate two excel sheets one with parallel threads enabled and other with disabled.
Each excel sheet has the system details as well as benchmarking numbers for gray image and RGB images with 1080p images

## Submission Checklist


rrawther (changes requested):
Generated from review-pr skill
 Critical

  1. rppt_sobel_filter called with wrong arguments (rpp_benchmarks.cpp:~451)

  rppt_sobel_filter(imgs[i].data, &srcDescs[i], out[i].data, &dstDescs[i],
                    sobelType, borderType, &rois[i], RpptRoiType::XYWH, handle, RPP_HOST_BACKEND);

  The rppt_sobel_filter API expects (sobelType, kernelSize, roiPtr, ...) but the code passes borderType (an RpptImageBorderType enum) where kernelSize should be. The
  borderType parameter doesn't exist in this API's signature. This will produce incorrect results — kernelSize will be whatever integer value RpptImageBorderType::REPLICATE
  maps to.

  2. rand() called inside #pragma omp parallel for — undefined behavior (multiple locations in both rpp_benchmarks.cpp and opencv_benchmarks.cpp)

  rand() uses global shared state and is not thread-safe. At least 6 benchmark functions in each of rpp_benchmarks.cpp and opencv_benchmarks.cpp call rand() from parallel OMP
   regions (Erase, CoarseDropout, RandomErase, GridDropout, CropAndPatch, CutoutDropout). This is a data race. Use std::mt19937 per-thread or thread_local engines instead —
  as is already done correctly in benchmark_OpenCV_ChannelDropout.

  3. RPP called per-image instead of batched — undermines benchmark validity (systemic in rpp_benchmarks.cpp)

  Every RPP benchmark sets desc.n = 1 and calls the API per-image in an OMP loop. RPP's tensor APIs are designed for batched processing (desc.n = batchSize) where the library
   handles parallelism internally. This:
  - Misses RPP's batch-level SIMD and threading optimizations
  - Adds artificial OMP overhead external to RPP
  - Makes the comparison misleading — it benchmarks "RPP per-image in user-managed parallelism" vs "OpenCV per-image"

  4. OpenCV internal threading not controlled (missing cv::setNumThreads())

  OpenCV operations like GaussianBlur, resize, warpAffine, etc. use internal thread pools. When ENABLE_PARALLEL_THREADS=ON with NUM_THREADS=128, the outer OMP loop spawns 128
   threads, each running internally-parallelized OpenCV calls — causing massive thread oversubscription (potentially 128 x N threads). When ENABLE_PARALLEL_THREADS=OFF, RPP
  gets 1 thread but OpenCV still uses its internal parallelism. Either way, the comparison is unfair. Add cv::setNumThreads() to match RPP's configuration.

  Major

  5. benchmark_RPP_BitwiseXor — potential buffer overread with mixed-size images (rpp_benchmarks.cpp:~1367)

  int idx2 = (i + 1) % num_images;
  rppt_bitwise_xor(imgs[idx1].data, imgs[idx2].data, &srcDescs[idx1], ...);

  Only one srcDesc is passed (for idx1) to describe both inputs. If imgs[idx1] and imgs[idx2] differ in size, this reads out-of-bounds memory. Contrast with
  BitwiseAnd/BitwiseOr which correctly copy images to ensure matching sizes.

  6. Multiple benchmarks assume all images have the same dimensions as imgs[0] (rpp_benchmarks.cpp)

  Transpose, Normalize, FusedMultiplyAddScalar, Slice, Remap, and LensCorrection all compute buffer sizes from imgs[0].rows/imgs[0].cols and use them for all images. With a
  non-uniform dataset, this causes buffer overflows or truncated data.

  7. benchmark_OpenCV_NoiseShot uses Vec3f unconditionally (opencv_benchmarks.cpp:~1158)

  float val = temp.at<Vec3f>(y, x)[c];

  This accesses memory as 3-channel even for grayscale images. Currently safe because NoiseShot is only called with RGB in main(), but this is a latent crash waiting for
  someone to add a grayscale call.

  8. benchmark_OpenCV_NonLinearBlend computes a different operation than RPP (opencv_benchmarks.cpp:~1319)

  The OpenCV version computes src1 * src2 / 255 (element-wise multiply). RPP's rppt_non_linear_blend uses a Gaussian-weighted blend with stdDev. These are fundamentally
  different operations, making this comparison meaningless.

  9. generate_test_dataset.py — pixel-by-pixel Python loops on 1080p images

  generate_noise_image and generate_radial_pattern_image use nested for y in range(HEIGHT): for x in range(WIDTH): img.putpixel(...) loops — 2M iterations each. This makes
  dataset generation take minutes instead of seconds. Use NumPy arrays: np.random.randint(0, 256, (1080, 1920, 3), dtype=np.uint8).

  Minor

  10. No RPP return status checking (entire rpp_benchmarks.cpp)

  Not a single RppStatus return value is checked. Failed API calls silently produce garbage output and report timing as if they succeeded.

  11. Inconsistent timing precision (both benchmark files)

  Most benchmarks use duration_cast<milliseconds> (integer truncation), but statistical operations use duration<double, milli>. Fast operations measured with integer
  precision could report 0 ms.

  12. benchmark_RPP_Threshold — thresh parameter unused (rpp_benchmarks.cpp:~1202)

  The function accepts double thresh but hardcodes minVal = 0.0f, maxVal = 255.0f. The thresh value only appears in the output string.

  13. benchmark_OpenCV_Remap — off-by-one in map_y (opencv_benchmarks.cpp:~801)

  map_y.at<float>(j, i) = map_x.rows - j;

  When j=0, this maps to rows (out of valid range 0..rows-1). Should be rows - 1 - j.

  14. toGenericDesc always maps dims as NCHW (benchmarks_utils.cpp)

  Sets dims[1] = desc.c regardless of layout. For NHWC data (which OpenCV Mats use), the physical dimension order is N,H,W,C. Whether this causes bugs depends on how RPP
  interprets dims vs strides, but it's at minimum misleading.

  15. generate_test_dataset.py — no random seed for reproducibility

  random.randint() is used without random.seed(). Each run produces different images, making cross-run benchmark comparisons non-reproducible.

  Nit

  16. README references nonexistent setup_and_build.sh — should be run_benchmarking.sh

  17. README uses internal path /path/to/rpp_ms_develop/ — should be a generic placeholder

  18. Dead benchmark_OpenCV_Emboss(imgs, isColor) 2-argument overload — superseded by the 4-parameter version

  19. Unused Python imports: sys and ImageFont are imported but never used

  20. Missing Pillow dependency in README manual instructions — only the automated script installs it

  ---
  Positive Observations

  - Comprehensive coverage: ~60 operations across color, filter, geometric, arithmetic, bitwise, statistical, effects, dropout, and composite categories
  - Clean separation of concerns: utilities, RPP benchmarks, OpenCV benchmarks, and main driver in separate files
  - Good Excel reporting with system info, per-operation results, and speedup ratios
  - The BenchmarkResult matching system (printResult with key-based pairing) is clever and handles operation variants (Resize modes, Flip directions) well
  - Shell script handles dependency installation, dataset generation, and dual-configuration builds automatically
  - dropout_helpers.cpp correctly uses std::mt19937 for thread-safe RNG

  ---
  Overall Assessment: Request Changes

  The benchmarking infrastructure is ambitious and well-structured, but has several issues that undermine the validity of the results:

  1. The comparison methodology is fundamentally flawed — RPP is called per-image (missing its batch optimizations) while OpenCV threading is uncontrolled. Results from this
  benchmark could be misleading in either direction.
  2. There are real bugs — wrong API arguments to rppt_sobel_filter, thread-unsafe rand() calls, potential buffer overflows with non-uniform image sizes.
  3. At least one operation comparison is invalid — NonLinearBlend computes entirely different math.

  Recommended fixes before merge (priority order):
  1. Fix the rppt_sobel_filter argument bug
  2. Replace all rand() in OMP regions with thread-safe RNG
  3. Add cv::setNumThreads() to control OpenCV's internal parallelism
  4. Consider restructuring RPP calls to use batched mode (desc.n = batchSize)
  5. Fix NonLinearBlend to compute the same operation or remove the comparison
  6. Add bounds checking for variable-size image batches
  7. Replace pixel-by-pixel Python loops with NumPy in the dataset generator

utilities/test_suite/OPENCV/benchmarks_common.h
```diff
@@ -0,0 +1,275 @@
...
+#include <sys/utsname.h>
+#include <iomanip>
+#include <map>
+#include <random>
+#include "rpp.h"
+#include "rpp_version.h"
+#include "xlsxwriter.h"
+
+// Test image paths
+#define GRAY_IMAGE_PATH "1080p_128images_dataset/"
```
LakshmiKumar23, line 56: Can we just pass this instead as a user parameter in the script?

utilities/test_suite/OPENCV/benchmarks_utils.cpp
```diff
@@ -0,0 +1,575 @@
...
+    roi3d.xyzwhdROI.roiDepth = 1;
+    return roi3d;
+}
+
+// ==================== RPP COLOR AUGMENTATIONS ====================
+void writeResultsToExcel(const string& filename,
+                        const vector<BenchmarkResult>& grayResults,
+                        const vector<BenchmarkResult>& colorResults)
+{
+    lxw_workbook  *workbook  = workbook_new(filename.c_str());
```
LakshmiKumar23, line 351: Do we need a separate library for this? Can we dump into csv directly using fstream instead to reduce dependency

utilities/test_suite/OPENCV/CMakeLists.txt
```diff
@@ -0,0 +1,144 @@
...
+
+# Find libxlsxwriter (for Excel output)
+find_library(XLSXWRITER_LIBRARY
+    NAMES xlsxwriter libxlsxwriter
+    PATHS /usr/lib /usr/local/lib ${ROCM_PATH}/lib
+    PATH_SUFFIXES lib lib64
+)
+
+# Find RPP library
+find_package(rpp 3.0.0 QUIET)
```
LakshmiKumar23, line 69: Are the `single_image` APIs available as part of `3.0.0`? Please check the version once. Latest ToT is `3.1.2`

utilities/test_suite/OPENCV/dropout_helpers.cpp
```diff
@@ -0,0 +1,109 @@
...
+            maskPtrTemp[channel] = keepDist(rng);
+            atLeastOne |= maskPtrTemp[channel];
+        }
+
+        if (!atLeastOne)
+            maskPtrTemp[rng() % channels] = 1;
+    }
+}
+
+void init_cutout_dropout(int batchSize, int maxBoxesPerImage, Rpp32u* numOfBoxes, RpptRoiLtrb* anchorBoxInfoTensor, RpptROIPtr roiTensorPtrSrc, int channels, int BitDepthTestMode, int seed, int dropoutType, void *colorBuffer)
```
LakshmiKumar23, line 54: replicated code from test_suite/HIP or HOST. Can we move these to `test_suite_common.h` and reuse? Hard to maintain duplicated code

LakshmiKumar23: And other similar init calls too

ammallya:
Imported to ROCm/rocm-libraries PR - 8175

